PR #6, opened 2024-05-17
Testing package versioning

Joao-L-S-Almeida:
It will add a version number for the tarballs created via `pip -m build`


CarlosGomes98 (commented):
This seems to just comment things out. Can they instead be deleted?

Joao-L-S-Almeida:
Yes.

Joao-L-S-Almeida:
Done.

PR #8, opened 2024-05-22
fix tutorial notebook not using strings for optimizer

CarlosGomes98:
The tutorial notebook still had a few paths that were not made general.

Additionally, added a line in the readme and docs about how to install GDAL, which is a core dependency but can be painful to install without conda.


CarlosGomes98:
Modified the notebook to use data available in the git repo + download an example image. This is a super simple toy example (dataset with only one sample) but can be run with no external setup.

Joao-L-S-Almeida (approved):

PR #19, opened 2024-06-07
Replacing the field `output_bands` with `input_bands` to avoid misunderstanding. 

Joao-L-S-Almeida:


Joao-L-S-Almeida:
It may not be the bet moment to execute this break change.

PR #29, opened 2024-06-28
Docs/pipx

Joao-L-S-Almeida:
Pipx as an alternative to pip.

### DIFF
--- a/tests/test_backbones.py
+++ b/tests/test_backbones.py
@@ -1,6 +1,7 @@
 import pytest
 import timm
 import torch
+import importlib
 
 import terratorch  # noqa: F401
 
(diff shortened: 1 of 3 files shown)


tests/test_backbones.py
```diff
@@ -49,7 +50,16 @@ def test_vit_models_accept_multitemporal(model_name, input_224_multitemporal):
...
+# Swin IS NOT on HuggingFace
+@pytest.mark.parametrize("model_name", ["prithvi_swin_L", "prithvi_swin_B"])
+def test_swin_instantiation(model_name):
+     base_module = "terratorch.models.backbones.prithvi_swin"
+     module = importlib.import_module(base_module)
+     model_class = getattr(module, model_name) 
 
+     model = model_class(pretrained=False, pretrained_bands=[0,1,2,3,4,5,6,7,8,9],
+                         bands=[1,2,3,4,5,6])
+
```
CarlosGomes98, line 62: is this test redundant with `test_can_create_backbones_from_timm`? I think both execute through the same code paths

CarlosGomes98 (commented):
Overall, I'm not entirely convinced of the need for pipx. From my understanding, this is something usually used for commands that the user wants available from all their environments to be used in general. e.g. something like `poetry`. 

Terratorch I believe is more of a dependency of a specific project. I can imagine, for instance, the case down the line where some projects require a certain version of terratorch, while others require a different version. In this case for a single user on his machine pipx would make life somewhat hard.

But I'm open to the fact I might be missing some good arguments for why its a good idea. Does it offer some advantage over pip for our case?

Joao-L-S-Almeida:
Pipx installs a Python package in a similar way APT does for Debian-based Linux and Brew for Mac, it means, it installs the dependencies in background and creates an executable or a symlink shared on PATH, which allows the user run it without needing to create an environment.  I think it could be useful for an user interested in just running fine-tuning experiments without needing to configure anything.

tests/test_backbones.py
Joao-L-S-Almeida: Yes. I'll remove it.

CarlosGomes98:
Ok, in the end it is one more alternative some users may prefer

CarlosGomes98 (changes requested):
After the change in `tests/test_backbones.py` is done will approve, LGTM

tests/test_backbones.py
Joao-L-S-Almeida: Done.

CarlosGomes98 (approved):

PR #32, opened 2024-07-02
Feature/refactor + dependabot

CarlosGomes98:
- Refactor directory structure to remove the unnecessary `src` directory
- Set up dependabot 
- Added missing copyright headers (that is the bulk of the changes)


CarlosGomes98:
After merging, dependabot version updates should be enabled in the repo settings

terratorch/datamodules/__init__.py
Joao-L-S-Almeida, line 1: These datamodules are being imported but not used, right ?

terratorch/datasets/__init__.py
Joao-L-S-Almeida, line 1: Same question asked for the other __init__ file.

.github/workflows/test.yaml
Joao-L-S-Almeida, line 1: In which circumstances we want to perform the tests for individual pushes ? As we have the rule of creating PR before sending pushes to `main` I thought it was already covered by the pull request instruction.

Joao-L-S-Almeida:
In general, how the massive renaming from src/terratorch to terratorch will affect the still open PR's  (specially https://github.com/IBM/terratorch/pull/19) ?

CarlosGomes98:
> In general, how the massive renaming from src/terratorch to terratorch will affect the still open PR's (specially #19) ?

@Joao-L-S-Almeida This is a good point. Should we hold off on this one until the currently open PRs are merged, and then do it? This should help minimize any headaches from it

.github/workflows/test.yaml
CarlosGomes98: Its a good point. Tbh I mostly took this from what I saw on TorchGeo. Could it be for the case where a force push to main happens?

terratorch/datamodules/__init__.py
CarlosGomes98: They are not being used, but they enable `from terratorch.datamodules import`, rather than needing to go down to the module

terratorch/datasets/__init__.py
CarlosGomes98: Yes, same thing.

.github/workflows/test.yaml
Joao-L-S-Almeida: Right. I believe it won't affect our current workflow, but it could cover the exceptional cases in which we need to direclty push to the master. I agree.

Joao-L-S-Almeida:
I believe just #19 may be affected by these changes. Maybe we could consider to approve or discard it before approving with the current PR. 
Since I have no more comments about this PR, I'll approve it as soon this question is decided.

Joao-L-S-Almeida (approved):

PR #33, opened 2024-07-03
Using the Lightning CLI interface to run tests

Joao-L-S-Almeida:
Executing the CLI tests using the Lightning API

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,4 +32,5 @@ jobs:
         run: pip list
       - name: Test with pytest
         run: |
+          export PYTHONPATH=.
           pytest -s tests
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -32,4 +32,5 @@ jobs:
         run: pip list
       - name: Test with pytest
         run: |
+          export PYTHONPATH=.
```
CarlosGomes98, line 35: This strikes me as a bit of a code smell. Why do we need it for these tests? The others seemed to be working without it

Joao-L-S-Almeida: Yes. It has no function here.

CarlosGomes98:
Mistakenly added my `--signoff` to your previous commits too, sorry

CarlosGomes98 (approved):

PR #44, opened 2024-07-08
Trying to test for 3.12 (update packages if necessary)

Joao-L-S-Almeida:


Joao-L-S-Almeida:
Seemingly the toolkit is working for python 3.10, 3.11 and 3.12 using torch == 2.3.*, considering the flexible versions defined on `requirements/required.txt`, however, perhaps we could condisider to use a discontinuous definition like   `torch<=2.1, >=2.3` if we observe some incompatibility.

CarlosGomes98:
these requirements files should now really only be updated through accepting PRs from dependabot, and they should always have exactly pinned versions. For requirements like `torch >= 2.1.` we should add this instead in the pyproject.toml

Joao-L-S-Almeida:
Maybe we could create a kind of conditional for the requirements when switching between versions.

Joao-L-S-Almeida:
I defined new fixed versions for the packages in `requirements/required.txt` based in the choices made automatically by Actions . It seems that is working for all the tests and python versions.

romeokienzler:
lgtm - fix conflicts..then merge

romeokienzler:
thanks @Joao-L-S-Almeida 
- [x] merged

PR #52, opened 2024-07-16
New unit tests

Joao-L-S-Almeida:
These tests were partially generated using Granite-20b.


CarlosGomes98 (commented):
I think we maybe forgot about this. lets close if we dont think these are necessary

PR #56, opened 2024-07-19
extends smp_model_factory class

PedroConrado:


Joao-L-S-Almeida:
Thanks for the contribution @PedroConrado.  Could you provide some basic unit tests just to check if your modules are not crashing when using the current terratorch requirements ? Could you provide the versions you need for the dependencies you have used (as segmentation-models-pytorch ) ?

terratorch/models/smp_model_factory.py
```diff
@@ -18,66 +99,222 @@ def build_model(
...
+            if backbone.startswith("prithvi"): # Using Prithvi encoder (ViT or Swin).
+                backbone_class = self._make_smp_encoder(PrithviModelWrapper)
+                if  backbone.startswith("prithvi_swin"):
+                    backbone_kwargs['prithvi_class'] = MMSegSwinTransformer
+                elif backbone.startswith("prithvi_vit"):
+                    backbone_kwargs['prithvi_class'] = TemporalViTEncoder
+                else:
+                    msg = f"Prithvi Backbone not found."
+                    raise NotImplementedError(msg)
+            # Using new encoder (not Prithvi or SMP).
```
CarlosGomes98, line 200: using `timm.create_model` should handle all these complexities. `checkpoint_path` should also be something that we don't take, but that is passed to this constructor (See the info box [here](https://ibm.github.io/terratorch/models/))

terratorch/models/smp_model_factory.py
```diff
@@ -18,66 +99,222 @@ def build_model(
...
+                self._out_channels = out_channels
+                
+                # If PFFIM is passed from config file
+                if "prepare_features_for_image_model" in kwargs:
+                    path = kwargs['prepare_features_for_image_model'] 
+                    self.prepare_features_for_image_model = _get_callable_from_path(path)  
+                # If PFFIM is in super                  
+                elif hasattr(super(), 'prepare_features_for_image_model'):
+                    self.prepare_features_for_image_model = super().prepare_features_for_image_model
+                # No PFFIM
```
CarlosGomes98, line 281: Shouldn't inheritance take care of this? why is it necessary?

terratorch/models/smp_model_factory.py
```diff
@@ -1,14 +1,95 @@
...
+import torch
 from torch import nn
 
+from terratorch.datasets import HLSBands
+from terratorch.models.backbones.prithvi_vit import checkpoint_filter_fn
 from terratorch.models.model import Model, ModelFactory, ModelOutput, register_factory
+from terratorch.models.backbones.vit_encoder_decoder import TemporalViTEncoder
+from terratorch.models.backbones.swin_encoder_decoder import MMSegSwinTransformer
+
+import torch.nn.functional as F  # noqa: N812
```
CarlosGomes98, line 17: Make sure imports are sorted (ruff) should do it

terratorch/models/smp_model_factory.py
```diff
@@ -1,14 +1,95 @@
...
+                    ).permute(0,3,1,2).contiguous()
+                    for layer_output in outs
+                ]
+
+            self.model.prepare_features_for_image_model = prepare_features_for_image_model
+        elif "TemporalViTEncoder" in self.prithvi_class:
+            self.model = prithvi_class(**kwargs, encoder_only=True)
+        else:
+            self.model = prithvi_class(**kwargs) 
+
```
CarlosGomes98, line 46: empty space

terratorch/models/smp_model_factory.py
```diff
@@ -1,14 +1,95 @@
...
+            self.model = prithvi_class(**kwargs, encoder_only=True)
+        else:
+            self.model = prithvi_class(**kwargs) 
+    
+    def forward(self, x):
+        return self.model.forward_features(x)
+
+    def prepare_features_for_image_model(self, x):
+        return self.model.prepare_features_for_image_model(x)
+
```
CarlosGomes98, line 52: empty space

terratorch/models/smp_model_factory.py
```diff
@@ -1,14 +1,95 @@
...
+    
+    def forward(self, x):
+        return self.model.forward_features(x)
+
+    def prepare_features_for_image_model(self, x):
+        return self.model.prepare_features_for_image_model(x)
+    
+    def channels(self):
+        return self.config["num_heads"]*[self.config["embed_dim"]]
+
```
CarlosGomes98, line 55: empty space

terratorch/models/smp_model_factory.py
```diff
@@ -18,66 +99,222 @@ def build_model(
...
+            ValueError: If the specified decoder is not supported by SMP.
+            Exception: If the specified task is not "segmentation" or "regression".
 
         Returns:
-            Model: SMP model wrapped in SMPModelWrapper.
+            nn.Module: A model instance wrapped in SMPModelWrapper configured according to the specified
+                    parameters and tasks.
         """
-        if task not in ["segmentation", "regression"]:
+        self.CPU_ONLY = not torch.cuda.is_available()
```
CarlosGomes98, line 143: This class shouldn't have the responsibility for doing this, lightning does it.

terratorch/models/smp_model_factory.py
```diff
@@ -18,66 +99,222 @@ def build_model(
...
+            # Can load model from local checkpoint. 
+            if "checkpoint_path" in backbone_kwargs:
+                
+                checkpoint_path = backbone_kwargs.pop("checkpoint_path")
+                print(f"Trying to load from the path defined in the config file, {checkpoint_path}.")
+
+                if self.CPU_ONLY:
+                    model_dict = torch.load(checkpoint_path, map_location="cpu")
+                else:
+                    model_dict = torch.load(checkpoint_path)
```
CarlosGomes98, line 189: We shouldn't need this because timm handles model instantiation, for for future reference, you can just always use line 187, even if the model will be loaded onto GPU :)

terratorch/models/__init__.py
```diff
@@ -6,6 +6,8 @@
 from terratorch.models.smp_model_factory import SMPModelFactory
 from terratorch.models.timm_model_factory import TimmModelFactory
 
+from terratorch.models.smp_model_factory import get_smp_decoder
```
CarlosGomes98, line 9: I dont think we should expose this directly. Is there a good reason for importing this here?

terratorch/models/__init__.py
```diff
@@ -15,4 +17,5 @@
     "TimmModelFactory",
     "AuxiliaryHead",
     "AuxiliaryHeadWithDecoderWithoutInstantiatedHead",
+    "get_smp_decoder",
```
CarlosGomes98, line 20: same as above

terratorch/models/__init__.py
```diff
@@ -6,6 +6,8 @@
 from terratorch.models.smp_model_factory import SMPModelFactory
 from terratorch.models.timm_model_factory import TimmModelFactory
```
CarlosGomes98, line 8: unsorted imports (ruff can sort imports)

terratorch/models/prithvi_model_factory.py
```diff
@@ -95,7 +98,13 @@ def build_model(
                 msg = f"Task {task} not supported. Please choose one of {SUPPORTED_TASKS}"
                 raise NotImplementedError(msg)
 
+            # These params are used in case we need a SMP decoder
+            # but should not be used for timm encoder
             backbone_kwargs, kwargs = _extract_prefix_keys(kwargs, "backbone_")
+            smp_kwargs, kwargs = _extract_prefix_keys(kwargs, "smp_")
+            aux_kwargs, kwargs = _extract_prefix_keys(kwargs, "aux_")
```
CarlosGomes98, line 105: Can we make use of the args passed through `decoder_` instead of needing a new prefix for smp and aux?

terratorch/models/smp_model_factory.py
```diff
@@ -1,83 +1,387 @@
...
+
+    @property
+    def output_embed_dim(self):
+        return self.channels
+
+    def forward_multiple_embeds(self, *x):
+        return self.decoder(*x)
+
+    def forward_single_embed(self, x):
+        return self.decoder(x[-1])
```
CarlosGomes98, line 49: Instead of this, can we remain consistent with current Prithvi decoders and accept an `in_index` argument? This should be a single int or a list of ints that are the indices of `x` which we select and pass forward to the decoder

terratorch/models/smp_model_factory.py
```diff
@@ -1,83 +1,387 @@
...
+                "output_stride": backbone_kwargs,
+            }
+
+            for param, config_dict in required_params.items():
+                if param not in config_dict:
+                    msg = f"Config must include the '{param}' parameter"
+                    raise ValueError(msg)
+
+            # Using new encoder.
+            backbone_class = self._make_smp_encoder(backbone)
```
CarlosGomes98, line 198: is this working? `_make_smp_encoder` seems to be a module level method rather than a method of this class

terratorch/models/smp_model_factory.py
```diff
@@ -1,83 +1,387 @@
...
+
+    # Creates model with dummy encoder and decoder.
+    model = decoder_module(**model_args, aux_params=aux_kwargs)
+
+    # Wrapper for SMP Decoder.
+    smp_decoder = SMPDecoderForPrithviWrapper(decoder=model.decoder, num_channels=out_channels[-1])
+    if "multiple_embed" in head_kwargs:
+        smp_decoder.forward = smp_decoder.forward_multiple_embeds
+    else:
+        smp_decoder.forward = smp_decoder.forward_single_embed
```
CarlosGomes98, line 306: see suggestion above to allow `in_index` parameter, which would make this monkey patching not necessary

terratorch/models/smp_model_factory.py
```diff
@@ -1,83 +1,387 @@
...
+def _register_custom_encoder(encoder, params, pretrained):
+    smp_encoders["SMPEncoderWrapperWithPFFIM"] = {
+        "encoder": encoder,
+        "params": params,
+        "pretrained_settings": pretrained
+    }
+
+
+# Gets class either from string or from Module reference.
+def _make_smp_encoder(encoder=None):
```
CarlosGomes98, line 321: should this be a method of the factory above?

terratorch/models/smp_model_factory.py
```diff
@@ -1,83 +1,387 @@
...
         )
 
 
-class SMPModelWrapper(Model, nn.Module):
-    def __init__(self, smp_model, relu=False, squeeze_single_class=False) -> None:
-        super().__init__()
-        self.smp_model = smp_model
-        self.final_act = nn.ReLU() if relu else nn.Identity()
-        self.squeeze_single_class = squeeze_single_class
+def get_smp_decoder(
```
CarlosGomes98, line 227: this function and the class `SMPDecoderForPrithviWrapper` seem to only be used for creating decoders which match the format required by `PrithviModelFactory` right? Then, would it make sense to instead move them to the `PrithviModelFactory` class?

PedroConrado: This function is also used by _get_smp_encoder to create the dummy smp encoder.

CarlosGomes98 (approved):
Looks good!

PR #58, opened 2024-07-24
Avoiding to remove info, but the ordering is not preserved.

Joao-L-S-Almeida:


terratorch/cli_tools.py
```diff
@@ -130,30 +130,31 @@ def write_on_epoch_end(self, trainer, pl_module, predictions, batch_indices):  #
                 save_prediction(prediction, file_name, output_dir, dtype=trainer.out_dtype)
 
 
-def clean_config_for_deployment_and_dump(config: dict[str, Any]):
+def clean_config_for_deployment_and_dump(config: dict[str, Any], clean:bool=False):
```
CarlosGomes98, line 133: weird spacing in the keyword arguments

CarlosGomes98:
I don't see any way for this `clean` argument to be passed to this function when the user is training, either with a yaml config file or programmatically. This could be passed in the config itself.

Joao-L-S-Almeida:
I transformed it into an input argument: dce51521bf6296b571029225dcf1e9718091ccb7
I believe that's right now.

Joao-L-S-Almeida:
@CarlosGomes98 What will be the destiny for this PR ? Should we close it without merging ?

Joao-L-S-Almeida:
Closing since #109 does it in a better way.

Issue #64, opened 2024-07-31
Albumentations > 1.4.10 breaks config parsing


CarlosGomes98:
Still not sure quite why this is happening, but albumentations > 1.4.10 seems to break the jsonargparse config parsing. All that happened from 1.4.11 was the base class for these augmentations had its parameter order changed.

For now will pin albumentations

Joao-L-S-Almeida:
We pinned albumentations to 1.4.0 and it seems to be sufficient for now.

PR #65, opened 2024-07-31
pin albumentations

CarlosGomes98:
see #64


terratorch/datasets/generic_pixel_wise_dataset.py
Joao-L-S-Almeida, line 1: It doesn't seem to related to the issue, but I agree with that.

terratorch/datasets/generic_scalar_label_dataset.py
Joao-L-S-Almeida, line 1: It doesn't seem to related to the issue, but I agree with that.

Joao-L-S-Almeida (approved):

PR #68, opened 2024-08-01
Allowing to overwrite Prithvi and new segmentation tests.

Joao-L-S-Almeida:
Segmentation was not being tested during fine-tuning tests.


Joao-L-S-Almeida:
@CarlosGomes98 @romeokienzler The conflicts were solved.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -139,13 +141,22 @@ def create_prithvi_vit_100(
...
         "num_frames": 1,
     }
+
+    # It is possible to overwrite default parameters using 
+    # config file
+    kwargs_ = defaultdict()
+    kwargs_.update(model_args)
+    kwargs_.update(kwargs)
+    kwargs_ = dict(kwargs_)
+
```
CarlosGomes98, line 151: why is this necessary? the previous line 147 should already cause `model_args` to be overwritten if the same key exists in kwargs. 

```python
>>> a = {"b": 2}
>>> kwargs = {"b": 3, "c": 4}
>>> dict(a, **kwargs)
{'b': 3, 'c': 4}
```

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -81,7 +83,7 @@ def _create_prithvi(
     if "features_only" in kwargs:
         kwargs = {k: v for k, v in kwargs.items() if k != "features_only"}
 
-    kwargs["in_chans"] = len(model_bands)
+    kwargs["in_chans"] = _estimate_in_chans(model_bands=model_bands)
```
CarlosGomes98, line 86: If the user passes `model_bands` with intervals, don't we need to "expand" this into the full form in order to use it in the `checkpoint_filter_fn`? I think this part is missing.

Then, once we have expanded, we should just be able to call `len` on that. I think we shouldn't need this `_estimate_in_chans`, the above approach should be simpler.

CarlosGomes98 (changes requested):
Adding the segmentation tests seems good to me. But the other two changes I see some issues with. Happy to discuss though

terratorch/models/backbones/prithvi_vit.py
Joao-L-S-Almeida: That's true. I don't know why it wasn't working for my test.

terratorch/models/backbones/prithvi_vit.py
Joao-L-S-Almeida: Maybe the modification from the last commits solves this issue.

terratorch/datasets/generic_pixel_wise_dataset.py
```diff
@@ -152,7 +152,7 @@ def __getitem__(self, index: int) -> dict[str, Any]:
         if self.expand_temporal_dimension:
             image = rearrange(image, "(channels time) h w -> channels time h w", channels=len(self.output_bands))
         image = np.moveaxis(image, 0, -1)
-
+
```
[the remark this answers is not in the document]
Joao-L-S-Almeida: About it and others related issues, I'll run the codebase under ruff and send it as as separated pull request.

terratorch/datasets/generic_pixel_wise_dataset.py
```diff
@@ -167,7 +167,7 @@ def __getitem__(self, index: int) -> dict[str, Any]:
         if self.transform:
             output = self.transform(**output)
         output["filename"] = self.image_files[index]
-
+
```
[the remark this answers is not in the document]
Joao-L-S-Almeida: About it and others related issues, I'll run the codebase under ruff and send it as as separated pull request.

terratorch/models/backbones/utils.py
```diff
@@ -0,0 +1,15 @@
+from terratorch.datasets import HLSBands
+
+def _are_sublists_of_int(item) -> (bool, bool):
```
CarlosGomes98, line 3: I believe this is never used. if so, can we remove it?

CarlosGomes98 (dismissed):
After removing that function (if it really is never used) lgtm. Lets do that ruff on the codebase after then :)

Joao-L-S-Almeida:
Done.

Joao-L-S-Almeida:

PR #71, opened 2024-08-02
Fix/mac local lambda

CarlosGomes98:
- Locally on mac, the previous unnamed lambda was not picklable, so tests were failing.
- The assertion that there are no duplicates breaks previous configs (including one on hf) that use duplicates on dataset bands (e.g. -1 on all bands that are irrelevant). Removed that check for now


Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Feel free to merge.

Joao-L-S-Almeida:
I don't see any problem in using 'auto', since Lightning have successfully detected gpus when available. However, I believe that setting 'cpu' is less error prone.

PR #74, opened 2024-08-02
adds smp model factory example

PedroConrado:


romeokienzler:
thanks @PedroConrado 

- [x] merged

PR #106, opened 2024-08-09
Extend/bands

Joao-L-S-Almeida:
This minor adjust will guarantee that the number of input channels during the backbone instantiation is properly estimated, even when the bands are defined as "lists of subintervals" with integers as limits. 
This is particularly useful when we have a large number of input bands, for example:
```
dataset_bands:
 - [0, 285]
 input_bands:
- [0, 50]
- [85, 150]
```


romeokienzler:
@CarlosGomes98 suggesting: infer no of in channels via band definitions

PR #109, opened 2024-08-12
Directly copying the YAML config to the logdir

Joao-L-S-Almeida:
This version guarantees that the input YAML file is directly (and exactly) copied to the log directory (for example, `<log_dir>/version_<number>/*.yaml`).


Joao-L-S-Almeida:
@romeokienzler and @CarlosGomes98 Maybe this is the shortcut to guarantee that the "outputted config file" is exaclty equal to the input one.

romeokienzler:
@Michal-Muszynski would this solution be sufficient for you?

Michal-Muszynski:
@romeokienzler having a copy of an input config file in the log directory (for example, <log_dir>/version_<number>/*.yaml) works well but there are two other output config files **config.yaml** and **config_deploy.yaml** created by terratorch that are not required to replicate an experiment in the future. Those two output config files (**config.yaml** and **config_deploy.yaml**) that are created based on the original input config file with some changes seem to be redundant.

Joao-L-S-Almeida:
Now you just need to add the a flag like:
` terratorch fit -c ~/Datasets/manufactured-finetune_prithvi_swin_B_segmentation.yaml --deploy_config_file False`
And those file will no more be generated.

Joao-L-S-Almeida:
@romeokienzler @Michal-Muszynski @CarlosGomes98 Are we going to merge this patch ?

terratorch/cli_tools.py
```diff
@@ -206,30 +214,36 @@ def setup(self, trainer: Trainer, pl_module: LightningModule, stage: str) -> Non
                 # save only on rank zero to avoid race conditions.
                 # the `log_dir` needs to be created as we rely on the logger to do it usually
                 # but it hasn't logged anything at this point
-                fs.makedirs(log_dir, exist_ok=True)
+                if self.deploy_config_file:
+                    fs.makedirs(log_dir, exist_ok=True)
```
romeokienzler, line 218: @Joao-L-S-Almeida  the log_dir is only created if the deploy_config_file is True - is that the intended semantics?

romeokienzler (commented):
added one question @Joao-L-S-Almeida

Joao-L-S-Almeida:
You are right, @romeokienzler . I modified it in the newest commit. The directory creation must be outside this conditional and it should just include the calls for `self.parser.save`

romeokienzler (approved):
thanks @Joao-L-S-Almeida lgtm

PR #110, opened 2024-08-13
set default optimizer to None

CarlosGomes98:
After a finetuning with a config file, there is currently a confusing key in the output config file, under the model optimizer.

This is caused by LightningCLI, which automatically overrides the `configure_optimizers` method of the task and instead instantiates the optimizers and schedulers based on the keys `optimizer` and `lr_scheduler` at the root level of the config file.

The confusion occurs due to the default values for the optimizer in the tasks, which get output in the config file nonetheless. Thus, in the config file, we end up with two different optimizers, even though the one under `model.optimizer` is not used.

To remove this confusion, this PR replaces the default value in the constructor with `None`. Then, in `configure_optimizers`, if the value is `None`, it replaces it with `Adam` (the current default value).

### DIFF
--- a/terratorch/tasks/segmentation_tasks.py
+++ b/terratorch/tasks/segmentation_tasks.py
@@ -46,7 +46,7 @@ class SemanticSegmentationTask(BaseTask):
         ignore_index: int | None = None,
         lr: float = 0.001,
         # the following are optional so CLI doesnt need to pass them
-        optimizer: str | None = "Adam",
+        optimizer: str | None = None,
         optimizer_hparams: dict | None = None,
         scheduler: str | None = None,
         scheduler_hparams: dict | None = None,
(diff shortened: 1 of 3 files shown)


terratorch/tasks/segmentation_tasks.py
Joao-L-S-Almeida, line 1: Is not that also necessary for classification tasks ?

Joao-L-S-Almeida (approved):

terratorch/tasks/segmentation_tasks.py
Joao-L-S-Almeida: @CarlosGomes98  When this last question is solved, feel free to merge it.

Joao-L-S-Almeida: I don't saw that. Maybe I definely became crazy.

PR #118, opened 2024-08-16
Removing hard-coded values when evaluating the number of patches. 

Joao-L-S-Almeida:
Number of patches must be evaluated using the value defined in the config not a hard-coded one.


CarlosGomes98:
Made some additions to also address #128, which is closely related. @Joao-L-S-Almeida the rest looks good to me. Please also give my changes a review :)

CarlosGomes98 (dismissed):

CarlosGomes98:
All good points!!

- We can remove this `pos_embed` altogether. We choose to compute it on the fly to allow for different image sizes to be passed to the same instance of the model. In this spirit, and together with your third point, we might actually want to get rid of this `num_frames` parameter altogether, and infer it at forward-time as well. From a quick look now, I don't think we actually need it for anything in order to instantiate the model. The `grid_size` and `num_patches` attributes of the `PatchEmbed` can then also be removed.
- I agree this can probably be made clearer. These dimensions we talk about don't actually exist in the transformer per-se, it just sees a big sequence of tokens. Rather they are important for the positional encoding step, and for when we do a reshape after the encoder. I'll do a rewrite and take your points into account.
- If the parameters of the patch embedding are different (patch size and tubelet size), I'm not sure we will be able to re-use any of the weights. In this situation, I suggest we check these parameters and, if they differ, initialize the patch embedding from scratch with a warning to the user. There might be a more optimal way to do it, but then this becomes more of a research question which I would discuss in a different issue / PR.

CarlosGomes98 (commented):
will integrate new features discussed above before merge

CarlosGomes98:
@Joao-L-S-Almeida as far as I can tell, having the weights saved and loaded causes the tests to be flakey on the github runner. Disabling this seems to make it work consistently. Not ideal, but something we might have to live with.

Unfortunately the message from the runner is not very helpful, I think it just dies

CarlosGomes98:
finetuning tests just seem to be randomly flakey. will remove for now.

CarlosGomes98 (approved):

PR #138, opened 2024-08-28
Fixed transform in generic classification dataset

blumenstiel:
The generic dataset resulted in an error because albumentations does not expect `label` and `filename` as an input.

### DIFF
--- a/terratorch/datamodules/generic_scalar_label_data_module.py
+++ b/terratorch/datamodules/generic_scalar_label_data_module.py
@@ -14,15 +14,14 @@ import torch
 from torch import Tensor
 from torch.utils.data import DataLoader
 from torchgeo.datamodules import NonGeoDataModule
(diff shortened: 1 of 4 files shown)


blumenstiel:
Also, `_generate_bands_intervals` was not present in the dataset.

CarlosGomes98:
@blumenstiel mostly added some cleanups piggy-backing on this PR. if you can check for your usecase as well and confirm it works we can merge

blumenstiel:
@CarlosGomes98 @Joao-L-S-Almeida I added
```
warnings.filterwarnings("ignore", category=rasterio.errors.NotGeoreferencedWarning)
```

to the NotGeo datasets, as we don't need a `geotransform` for these datasets.

CarlosGomes98 (approved):

PR #139, opened 2024-08-28
reorganize the tests directory

CarlosGomes98:


Joao-L-S-Almeida:
@CarlosGomes98 The config files were moved, but it's also necessary to update the relative path for the datasets and the metrics files (means.txt and stds.txt) for ach of them.

CarlosGomes98:
@Joao-L-S-Almeida thanks for spotting that! Should be ok now :)

Joao-L-S-Almeida (approved):

Issue #137, opened 2024-08-28
pip install git+https://github.com/IBM/terratorch.git not working


blumenstiel:
Building from GitHub is not working.
```
(venv) [blu@cccxc418 gfm-workshop]$ pip install git+https://github.com/IBM/terratorch.git
...
Building wheels for collected packages: terratorch
  Building wheel for terratorch (pyproject.toml) ... done
  Created wheel for terratorch: filename=terratorch-0.99.1-py3-none-any.whl size=7126 sha256=91a66402ea8e7410ea9a6a939c280e297472cfec7a2f86607ed38d73579c8de5
  Stored in directory: /tmp/pip-ephem-wheel-cache-slhwgpt3/wheels/af/36/d2/87c95a959b1674c314cfc1f375fa12875794262f65a60afbfe
Successfully built terratorch
Installing collected packages: terratorch
Successfully installed terratorch-0.99.1
(venv) [blu@cccxc418 gfm-workshop]$ python
Python 3.11.5 (main, Sep 22 2023, 15:34:29) [GCC 8.5.0 20210514 (Red Hat 8.5.0-20)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import terratorch
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'terratorch'
```

I checked the side packages and only `terratorch-0.99.1.dist-info` is added, but not the package itself.

PR #140, opened 2024-08-28
Fix/pyproject toml

CarlosGomes98:
#137 

Also took the chance to update and correct paths in the pyproject toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,8 +2,8 @@
 requires = [ "setuptools" ]
 build-backend = 'setuptools.build_meta'
 
-[tool.setuptools]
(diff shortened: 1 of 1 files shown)


CarlosGomes98:
@Joao-L-S-Almeida I intorduced `[tool.setuptools.packages.find]` instead of the previous solution with

```
[tool.setuptools]
py-modules = []
```

This still worked for me with `pip install -e .`. Could you confirm if this still works for your case?

blumenstiel:
Works for me, thanks @CarlosGomes98!

Joao-L-S-Almeida (approved):

PR #145, opened 2024-09-04
Extend/backbones decoders

Joao-L-S-Almeida:
* The factory to manage mmseg models was extended to support builtin versions of them. 
* First version of ASPPHead for regression and segmentation. 
* First version of U-Net for regression and segmentation.


Joao-L-S-Almeida:
@Michal-Muszynski 
An important question about this inclusion: are we going to possibly use ASPPHead as a single model or it will always be combined to an U-Net as a decoder head ?

Joao-L-S-Almeida:
@CarlosGomes98 @romeokienzler @Michal-Muszynski  I consider it as good enough to be merged. 
My questions are:
- [] The current suported configs are good enough to replace  mmseg for the same tasks ?
- [] Are we interested in including support for more decoders than ASPP for now ?

romeokienzler (approved):
@Joao-L-S-Almeida lgtm, great work, tnx a lot, pse merge!

PR #151, opened 2024-09-10
Trying to upgrade torchgeo

Joao-L-S-Almeida:
torchgeo >= 0.6.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 dependencies = [
   "torch>=2.1.0",
   "torchvision>=0.16.0",
-  "torchgeo>=0.5.1",
+  "torchgeo>=0.6.0",
   "rioxarray>=0.15.0",
   # see issue #64
   "albumentations>=1.3.1, <=1.4.10",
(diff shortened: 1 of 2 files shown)


CarlosGomes98:
The changes on requirements.txt are fine. But pyproject should try to be as lenient as possible so that `terratorch` plays nicely with as many libraries as possible. If a library out there requires torchgeo-0.5.2 for some reason, we would still like to be compatible with them. Basically it should have the lowest version we tolerate in order for all the functionalities to work. We should only change this part if we introduce a feature that really requires the newest torchgeo.

Like this, when someone installs terratorch, `pip` will install the newest possible `torchgeo` that is still compatible with all the other requirements the user has.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 dependencies = [
   "torch>=2.1.0",
   "torchvision>=0.16.0",
-  "torchgeo>=0.5.1",
+  "torchgeo>=0.6.0",
```
CarlosGomes98, line 34:
```suggestion
  "torchgeo>=0.5.1",
```

CarlosGomes98 (commented):
see above

Joao-L-S-Almeida:
@CarlosGomes98 Done.

CarlosGomes98 (approved):

PR #157, opened 2024-09-13
Make mmseg installation fully optional

Joao-L-S-Almeida:
It avoids problems with mmseg and its dependencies.


CarlosGomes98 (approved):

CarlosGomes98:
This is good to fix any installation issues for now, but we may be able to handle it more nicely with a separate requirements file in the future

PR #158, opened 2024-09-15
Adds temporal Datasets

PedroConrado:
adds PASTIS, Sen4AgriNet and OpenSentinelMap datasets and datamodules.

### DIFF
--- a/terratorch/datasets/open_sentinel_map.py
+++ b/terratorch/datasets/open_sentinel_map.py
@@ -16,6 +16,7 @@ from torchgeo.datasets import NonGeoDataset
 
 from terratorch.datasets.utils import pad_numpy, to_tensor
 
+MAX_TEMPORAL_IMAGE_SIZE = (192, 192)
 
 class OpenSentinelMap(NonGeoDataset):
     def __init__(
@@ -29,7 +30,6 @@ class OpenSentinelMap(NonGeoDataset):
         truncate_image: int | None = None,
         target: int = 0,
         pick_random_pair: bool = True,  # noqa: FBT002, FBT001
-        output_size: tuple[int, int] = (192, 192),
     ) -> None:
         """
         Pytorch Dataset class to load samples from the OpenSentinelMap dataset, supporting
@@ -48,8 +48,6 @@ class OpenSentinelMap(NonGeoDataset):
                 If None, no truncation is performed.
             target (int): Specifies which target class to use from the mask. Default is 0.
             pick_random_pair (bool): If True, selects two random images from the temporal sequence. Default is True.
-            output_size (tuple of int, optional): Desired output size (H, W) for the interpolated images.
-                Only used if spatial_interpolate_and_stack_temporally is True. Default is (192, 192).
         """
         split = "test"
         if bands is None:
@@ -81,7 +79,6 @@ class OpenSentinelMap(NonGeoDataset):
         self.truncate_image = truncate_image
         self.target = target
         self.pick_random_pair = pick_random_pair
-        self.output_size = output_size
 
         self.image_files = []
         self.label_files = []
@@ -182,7 +179,7 @@ class OpenSentinelMap(NonGeoDataset):
                     band_frame = torch.from_numpy(band_frame).float()
                     band_frame = band_frame.permute(2, 0, 1)
                     interpolated = F.interpolate(
-                        band_frame.unsqueeze(0), size=self.output_size, mode="bilinear", align_corners=False
+                        band_frame.unsqueeze(0), size=MAX_TEMPORAL_IMAGE_SIZE, mode="bilinear", align_corners=False
                     ).squeeze(0)
                     interpolated_bands.append(interpolated)
                 concatenated_bands = torch.cat(interpolated_bands, dim=0)
--- a/terratorch/datasets/open_sentinel_map.py
+++ b/terratorch/datasets/open_sentinel_map.py
@@ -14,7 +14,7 @@ from PIL import Image
 from torch import Tensor
 from torchgeo.datasets import NonGeoDataset
 
-from terratorch.datasets.utils import to_tensor, pad_numpy
+from terratorch.datasets.utils import pad_numpy, to_tensor
 
 
 class OpenSentinelMap(NonGeoDataset):
@@ -24,11 +24,12 @@ class OpenSentinelMap(NonGeoDataset):
         split: str = "train",
         bands: list[str] | None = None,
         transform: A.Compose | None = None,
-        interpolate_and_concat_bands: bool = True,  # noqa: FBT001, FBT002
+        spatial_interpolate_and_stack_temporally: bool = True,  # noqa: FBT001, FBT002
         pad_image: int | None = None,
         truncate_image: int | None = None,
         target: int = 0,
         pick_random_pair: bool = True,  # noqa: FBT002, FBT001
+        output_size: tuple[int, int] = (192, 192),
     ) -> None:
         """
         Pytorch Dataset class to load samples from the OpenSentinelMap dataset, supporting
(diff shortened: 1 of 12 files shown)


Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
I agree with these additions. Have you any observation @CarlosGomes98 ?

CarlosGomes98:
This is quite a large PR, so I'd like to give it a review as well. I should be able to by the end of the week / early next week. If possible let's hold off on merging until then

Joao-L-S-Almeida:
As I could verify, all the modifications are additions which don't conflict with the current main version. @PedroConrado  Maybe it could be interesting to include some basic tests under `tests/ `just to check if the datasets are working as you expect. 
We, in fact, should add these tests for all the datasets and datamodules currently deployed in the repository.

terratorch/datasets/pastis.py
```diff
@@ -0,0 +1,312 @@
...
+        data_root,
+        norm=True,  # noqa: FBT002
+        target="semantic",
+        folds=None,
+        reference_date="2018-09-01",
+        class_mapping=None,
+        transform = None,
+        truncate_image = None,
+        pad_image = None,
+        sats=["S2"],  # noqa: B006
```
CarlosGomes98, line 28: I would maybe prefer `satellites` for full clarity in naming

terratorch/datasets/pastis.py
```diff
@@ -0,0 +1,312 @@
...
+        self.truncate_image = truncate_image
+        self.pad_image = pad_image
+        # loads patches metadata
+        self.meta_patch = gpd.read_file(os.path.join(data_root, "metadata.geojson"))
+        self.meta_patch.index = self.meta_patch["ID_PATCH"].astype(int)
+        self.meta_patch.sort_index(inplace=True)
+
+        # stores table for each satalite date
+        self.date_tables = {s: None for s in sats}
+        self.date_range = np.array(range(-200, 600)) # date interval
```
CarlosGomes98, line 105: Any specific reason for this range? A comment to clarify might be good.

terratorch/datasets/sen4agrinet.py
```diff
@@ -0,0 +1,249 @@
+import random
```
CarlosGomes98, line 1: unsorted imports

terratorch/datasets/sen4agrinet.py
```diff
@@ -0,0 +1,249 @@
...
+        elif split == "test":
+            self.image_files = self.test_files
+
+    def __len__(self):
+        return len(self.image_files)
+
+    def split_data(self):
+        random.seed(self.seed)
+
+        if self.scenario == "random":
```
CarlosGomes98, line 94: With the seed fixed, this should produce the same sorting every time, but I would check this to be sure, as lightning sets seeds somewhat inconsistently some times. I'm almost sure its fine, but if its not, it could lead to issues. Could I ask you to just manually confirm that indeed the same split is produced if the key is the same?

terratorch/datasets/sen4agrinet.py
```diff
@@ -0,0 +1,249 @@
...
+            output = {}
+            images_over_time = []
+            for band in self.bands:
+                band_group = patch_data[band]
+                band_data = band_group[f"{band}"][:]
+
+                band_data = band_data.astype(np.float32)
+
+                if self.truncate_image:
+                    band_data = band_data[-self.truncate_image:]
```
CarlosGomes98, line 141: are they already sorted in time, such that we know we are getting only the most recent `self.truncate_image` images?

terratorch/datasets/sen4agrinet.py
```diff
@@ -0,0 +1,249 @@
...
+                if self.pad_image:
+                    band_data = pad_numpy(band_data, self.pad_image)
+
+                if self.interpolate_and_concat_bands:
+                    band_data = torch.from_numpy(band_data)
+                    band_data = band_data.clone().detach()
+
+                    interpolated = F.interpolate(
+                        band_data.unsqueeze(0), size=(366, 366), mode="bilinear", align_corners=False
```
CarlosGomes98, line 150: same comment as in the other dataset, is 366 just the size of the largest of all these images? And would it make sense to make this configurable? (maybe it doesnt, curious about your opinion)

terratorch/datasets/sen4agrinet.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if not all(band in sample for band in rgb_bands):
+            warnings.warn("No RGB image.")  # noqa: B028
+            return None
+
+        rgb_images = []
+        for t in range(sample["B04"].shape[0]):
+            rgb_image = torch.stack([sample[band][t] for band in rgb_bands])
+
+            # Normalização
```
CarlosGomes98, line 201: Maybe make this english ;)

CarlosGomes98 (commented):
Looks overall good to me, just a few comments / questions. Would be great to have tests to make sure they are working, even if simple.

terratorch/datasets/open_sentinel_map.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        Pytorch Dataset class to load samples from the OpenSentinelMap dataset, supporting
+        multiple bands and temporal sampling strategies.
+
+        Args:
+            data_root (str): Path to the root directory of the dataset.
+            split (str): Dataset split to load. Options are 'train', 'val', or 'test'. Defaults to 'train'.
+            bands (list of str, optional): List of band names to load. Defaults to ['gsd_10', 'gsd_20', 'gsd_60'].
+            transform (albumentations.Compose, optional): Albumentations transformations to apply to the data.
+            interpolate_and_concat_bands (bool): If True, the bands are interpolated and concatenated over time. Default is True.
```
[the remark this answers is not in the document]
PedroConrado: Yes, we interpolate the bands spatially (since they have different resolutions) so we can concatenate in the channels axis. after that, the images are temporally stacked. I wiil change it from "interpolate_and_concat_bands" to "spatial_interpolate_and_stack_temporally".

terratorch/datasets/open_sentinel_map.py
```diff
@@ -0,0 +1,222 @@
...
+            images_over_time = []
+            for _, npz_file in enumerate(npz_files):
+                data = np.load(npz_file)
+                interpolated_bands = []
+                for band in self.bands:
+                    band_frame = data[band]
+                    band_frame = torch.from_numpy(band_frame).float()
+                    band_frame = band_frame.permute(2, 0, 1)
+                    interpolated = F.interpolate(
+                        band_frame.unsqueeze(0), size=(192, 192), mode="bilinear", align_corners=False
```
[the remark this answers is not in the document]
PedroConrado: I think it does. The idea here was to interpolate it to the largest image and lose as little information as possible. But I guess we should leave that up to the user.

terratorch/datasets/sen4agrinet.py
PedroConrado: I checked manually with the following code and they were equal:

`
        split_hash = self._get_split_hash(split)
        previous_hash = self._load_split_hash()

        if previous_hash is None:
            self._save_split_hash(split_hash)
            print("First")
        elif split_hash == previous_hash:
            print("Equals")
        else:
            print("Diff")
`

terratorch/datasets/pastis.py
PedroConrado: This is the time range that was used in the paper. I will clarify that.

terratorch/datasets/open_sentinel_map.py
```diff
@@ -0,0 +1,222 @@
...
+                    band_images = [pad_numpy(img, self.pad_image) for img in band_images]
+                band_images = np.stack(band_images, axis=0)
+                final_image_dict[band] = band_images
+
+            output["image"] = final_image_dict
+
+        label_file = self.label_files[index]
+        mask = np.array(Image.open(label_file)).astype(int)
+        mask[mask == 254] = 15  # noqa: PLR2004
+        mask[mask == 255] = 16  # noqa: PLR2004
```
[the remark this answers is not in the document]
PedroConrado: This was done because the _unlabel_ and _none_ categories are represented by 254 and 255 in the mask, while the other categories range from 0 to 14. These categories were likely set as 254 and 255 to be used as ignore_index, but in other cases, it might be useful to include them. That's why I mapped them to unused category indices. Will clarify this further in the comments.

terratorch/datasets/sen4agrinet.py
PedroConrado: I made sure that they are indeed sorted now.

terratorch/datasets/open_sentinel_map.py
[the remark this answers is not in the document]
CarlosGomes98: Ah ok, that makes sense. Then I would leave it as this but make this 192 a constant defined at the top of the module with a meaningful variable name. If the user wants to resize, they should just use a transform.

CarlosGomes98 (approved):

PedroConrado:
Thanks for the review, maybe @Joao-L-S-Almeida or @CarlosGomes98 can merge? I dont have write access

Joao-L-S-Almeida (approved):

PR #165, opened 2024-09-24
bump to albumentations-1.4.16

romeokienzler:


blumenstiel:
@romeokienzler In my experience, this will break `terratorch fit` with Prithvi. Did you tested it?

CarlosGomes98:
+1. There were some changes in albumentations that break it's usage in the generic datasets / data modules when instantiating from a config file. Bumping it will require code modifications on our side.


________________________________

PR #170, opened 2024-09-24
Automatically setting batch_size to 1 when the images don't have all the same dimensions

Joao-L-S-Almeida:
When a a list of images with different dimensions is passed to a dataloader, the batch is automatically set to 1 to avoid stacking issues.


Joao-L-S-Almeida:

PR #177, opened 2024-09-30
remove geobench, benchmark script

CarlosGomes98:
- Geobench was never actually imported, and it imposes strict limitations in its dependencies
- Except in this benchmark script, which I've removed at least for now, as it is not super polished


CarlosGomes98:
@fmartiescofet

Joao-L-S-Almeida:
I agree. The last sequence of tests are failing, but GH runner is mad.

Joao-L-S-Almeida (approved):

PR #179, opened 2024-09-30
Extend/wxc

Joao-L-S-Almeida:


Joao-L-S-Almeida:
The fine-tuning already works for the programmatic interface. See `terratorch/examples/scripts/WxCTrain.py`

Joao-L-S-Almeida:

terratorch/datamodules/merra2_downscale.py
```diff
@@ -14,10 +14,10 @@ def __init__(self, split : str, **kwargs):
     def __getitem__(self, index) -> dict[Tensor | int]:
         batch = super().__getitem__(index)
         batch_extended = {}
+        batch_extended['mask'] = batch.pop("y")
```
romeokienzler, line 17: @Joao-L-S-Almeida why r u popping y?

terratorch/models/wxc_model_factory.py
```diff
@@ -43,4 +48,5 @@ def build_model(
         **kwargs,
     ) -> Model:
         module = get_finetune_model(kwargs['model_config'])
+        print("Using wrapper.")
```
romeokienzler, line 51: @Joao-L-S-Almeida can we use logging here?

terratorch/tasks/wxc_downscaling_task.py
```diff
@@ -42,9 +44,9 @@ def configure_models(self) -> None:
         self.model: Model = self.model_factory.build_model(
             "regression", aux_decoders=None, model_config=self.model_config, **self.hparams["model_args"]
         )
-        if self.hparams["freeze_backbone"]:
+        if self.hparams_["freeze_backbone"]:
```
romeokienzler, line 47: @Joao-L-S-Almeida what's the difference between hparams and hparams_?

romeokienzler: @CarlosGomes98 cc maybe we should come up with a better name for hparams_ which tells us the difference

terratorch/tasks/wxc_downscaling_task.py
```diff
@@ -69,22 +71,25 @@ def configure_losses(self) -> None:
...
         if loss == "mse":
-            self.criterion: nn.Module = nn.MSELoss(reduction="none")
+            self.criterion: nn.Module = nn.MSELoss(reduction="mean")
         elif loss == "mae":
-            self.criterion = nn.L1Loss(reduction="none")
+            self.criterion = nn.L1Loss(reduction="mean")
         elif loss == "rmse":
             # IMPORTANT! Root is done only after ignore index! Otherwise the mean taken is incorrect
-            self.criterion = RootLossWrapper(nn.MSELoss(reduction="none"), reduction=None)
+            self.criterion = RootLossWrapper(nn.MSELoss(reduction="none"), reduction="mean")
```
romeokienzler, line 84: @Joao-L-S-Almeida what was the reason to change this? would we need to expose this as parameter eventually? just making sure we don't hard code unexpected default behaviour...

romeokienzler:
@Joao-L-S-Almeida generally lvgtm! thanks so much! added 3-4 comments, in case u wanna have a quick look?

terratorch/models/wxc_model_factory.py
Joao-L-S-Almeida: Yes, see commit [b898ac3](https://github.com/IBM/terratorch/pull/179/commits/b898ac368add336fa08f5d93b9d7c944e671521c)

terratorch/tasks/wxc_downscaling_task.py
Joao-L-S-Almeida: I defined `self.hparams_` to be a dictionary containing all the original model config, since `self.hparams` (which I understand is being magically created by Lightning behind the scenes), is incomplete for my purposes. 
I renamed it to `self.extended_hparams` in commit [384eae7](https://github.com/IBM/terratorch/pull/179/commits/384eae789cea0e8e9f4db4ae6e0199dd8365e0ae).

terratorch/tasks/wxc_downscaling_task.py
Joao-L-S-Almeida: My bad, this choice is not necessary. I fixed it in the last commit.

examples/scripts/WxCTrain.py
```diff
@@ -0,0 +1,204 @@
...
+# ## Climatology
+# 
+# The PrithviWxC model was trained to calculate the output by producing a perturbation to the climatology at the target time. This mode of operation is set via the residual=climate option. This was chosen as climatology is typically a strong prior for long-range prediction. When using the residual=climate option, we have to provide the dataloader with the path of the climatology data.
+# 
+
+# 
+
+files = glob.glob("climatology/*")
+
+if not len(files):
```
CarlosGomes98, line 92: equivalent to `if not files`?

pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 ]
 
 dependencies = [
-  "torch>=2.1.0",
+  "torch>=2.1.0, <=2.5.0",
```
CarlosGomes98, line 29: does this break with newer torches?

terratorch/datamodules/merra2_downscale.py
```diff
@@ -30,8 +30,22 @@ def _dataloader_factory(self, split: str) -> DataLoader[dict[str, Tensor]]:
         return super()._dataloader_factory(split)
     
     def setup(self, stage: str) -> None:
+
+        if (stage == 'train'):
```
CarlosGomes98, line 34: no need for brackets around condition

terratorch/models/wxc_model_factory.py
```diff
@@ -43,4 +52,5 @@ def build_model(
         **kwargs,
     ) -> Model:
         module = get_finetune_model(kwargs['model_config'])
+        logger.info("Using wrapper.")
```
CarlosGomes98, line 55: is this necessary / helpful? it will always be printed, right?

terratorch/tasks/wxc_downscaling_task.py
```diff
@@ -31,6 +31,8 @@ def __init__(
 
         self.model_factory = get_factory(model_factory)
         self.model_config = model_config
+        # TODO Unify it with self.hparams
+        self.extended_hparams = self.model_config.to_dict()
```
CarlosGomes98, line 35: not quite sure why this is needed. if we want to add these to hparams, can we do `self.hparams |= self.model_config.to_dict()`?

terratorch/tasks/wxc_downscaling_task.py
```diff
@@ -97,20 +102,48 @@ def instantiate_metrics():
...
+        loss = self.train_loss_handler.compute_loss(model_output, y, self.criterion, None)
+        self.train_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=y.shape[0])
+        y_hat = model_output.output
+        self.train_metrics(y_hat, y)
+        self.log_dict(self.train_metrics, on_epoch=True)
+
+        return loss["loss"]
 
     def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None:
         raise NotImplementedError("This function is not yet implemented.")
```
CarlosGomes98, line 139: this should be pretty straightforward to add, given the others are already there

terratorch/tasks/wxc_downscaling_task.py
```diff
@@ -97,20 +102,48 @@ def instantiate_metrics():
...
 
     def predict_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> Tensor:
         x = batch["image"]
         file_names = batch["filename"]
-
+        
         def model_forward(x):
             return self(x).output
```
CarlosGomes98, line 147: unused function (I think this is the case in other tasks too, but there is no need for it here)

CarlosGomes98 (commented):
a few comments from my side, but overall lgtm

examples/scripts/WxCTrain.py
Joao-L-S-Almeida: Yes.

pyproject.toml
Joao-L-S-Almeida: It would break, but I believe the commit [1316272](https://github.com/IBM/terratorch/pull/179/commits/13162727f20219185733ec1705fe5bc25c2982ae) will solve that.

terratorch/models/wxc_model_factory.py
Joao-L-S-Almeida: It was just useful during the debugging. I'll remove it.

terratorch/tasks/wxc_downscaling_task.py
Joao-L-S-Almeida: I'm not sure about this comment.

terratorch/tasks/wxc_downscaling_task.py
Joao-L-S-Almeida: We could, but `self.hparams` is a `lightning.fabric.utilities.data.AttributeDict instance`.

Joao-L-S-Almeida:
Is it resolved ?

PR #180, opened 2024-10-03
add back geobench dependency

CarlosGomes98:
At least one of the geobench datasets sneakily uses geobench with `ast.literal_eval`. This snuck by because there is no `import` statement...

Added it back, and placed it in a new requirements/optional.txt


requirements/optional.txt
Joao-L-S-Almeida, line 1: Good. It's better to place all these dependencies in a separated file.

Joao-L-S-Almeida (approved):

PR #181, opened 2024-10-03
Added SMP and Timm examples

blumenstiel:
Added notebooks with one segmentation example (SMP) and one classification example (Timm)

@CarlosGomes98 Please let me know if you update the SMP or Timm factory so I can update the examples.


Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
The last pipeline of testes are being interrupted without any clear reason, but I believe it's a GH-related problem.

Issue #135, opened 2024-08-27
Moving towards a single model factory


paolofraccaro:
**Is your feature request related to a problem? Please describe.**
It is suboptimal and probably not scalable to have many different "model factories" for each model or libraries. 

**Describe the solution you'd like**
It would be great to figure out a solution where we abstract away (probably through some wrapping) the different libraries to contribute a single model factory. This is probably not straightforward but it will considerably improve the UX for terratorch.

paolofraccaro:
This makes sense to me. Let's try to think of this by allowing some flexibility in how we combine encoders withe decoders, especially for plugging in decoders this is important. 

The measures you propose seem sufficient to handle vast majority of cases. Passing functions to adapt encoder output to decoder input makes sense. I would extend this (maybe we can provide reshape transforms available as well) to adapt the output of a datamodule to encoder. 

Only question, do we do any check on the funxtions provided by the user? Thinking in case the code is submitted for execution somewhere else.

CarlosGomes98:
> Only question, do we do any check on the funxtions provided by the user? Thinking in case the code is submitted for execution somewhere else.

This is a good point, however, the user can already execute arbitrary datasets / datamodules / models, and all of those can execute arbitrary code as long as the user has access to the codebase. Without access to the codebase, the most the user could do is point to some function defined or imported somewhere. We might want to think about how to restrict this, but this is not the only mechanism that would need to be considered.

CarlosGomes98:
With this refactoring, it probably makes sense to do some refactoring on the tasks too. 

For instance, currently, they take the argument `aux_heads`, which should probably be passed directly to the factory - the task doesn't need this information

Joao-L-S-Almeida:
@CarlosGomes98 already did and deep and thorough evaluation about the issue, so I have just a few points to add (and maybe summarize what was exposed). 
- Our current decoder registry is a simple function which gets the corresponding class from a specific module (where all the possible decoder are pre-imported). Do you suggest to have a similar mechanism for the backbones, @CarlosGomes98 I agree that it would facilitate the instantiation and creation of new built-in backbones. We could even create a a kind of try-except constructor to search using timm and our inner registry. 

- The main problem with an universal class is that it aims to connect any pair (encoder, decoder) from any source and it is a non-trivial task, since the outputs produced by a particular encoder could be incompatible with the decoder chosen for it. 
[...]

PR #186, opened 2024-10-11
Feature/encoder decoder factory

CarlosGomes98:
This pr addresses #135 

It does quite a few things:
- Adds a more powerful EncoderDecoderFactory
- Adds registries and multisource registries
- Adds Necks
- Adds smp registry (@PedroConrado )
- Adds mmseg registry (@paolofraccaro )
- Updates examples to use the new functionalities
- Adds tests for these functionalities
- Adds docs for these functionalities

As future steps after this is merged, a lot of existing factories can either be removed / deprecated (as with PrithviModelFactory) or simplified greatly.

This PR does not break PrithviModelFactory, but modifies it to use the EncoderDecoderFactory. It is marked as deprecated and should be removed in the future. But for now this should not break any back compatibility.


pyproject.toml
Joao-L-S-Almeida, line 1: Why the preference for downloading geobench from GitHub ? I have observed issues with PyPI when installing in this way (which have me obligated to recommend the installation from tarballs for `wxc`).

requirements/optional.txt
Joao-L-S-Almeida, line 1: The same as commented above.

terratorch/models/prithvi_model_factory.py
Joao-L-S-Almeida, line 1: I understand the motivation behind these modifications and I agree that they are important to simplify the source code, however, mostly of this structure is still required for the pretraining PR, so they were partially reintroduced in the PR #193. 
Nevertheless, I believe we can go with them for now and just solve the conflicts with more patience for the pretraining tasks.

terratorch/models/generic_unet_model_factory.py
Joao-L-S-Almeida, line 1: It will maybe require another conflicting solver for the still open PR #145, but it can be done as a next step.

Joao-L-S-Almeida:
@CarlosGomes98 @romeokienzler 
General comment. I reviewed the new way to define factories and registries. I believe it simplifies the entire process and introduces clarification for some parts (specially for the registries). Even though it creates some conflicts for two open PRs, I believe it's easier to solve these conflicts when these modifications are already in the main branch. Especially considering that the pretraining and mmseg PRs are not so urgent.

pyproject.toml
CarlosGomes98: We submitted a PR for geobench for its dependencies to be loosened, as this was creating issues in some projects (for instance their version of huggingface-hub was an old one). This pr was accepted, however the release was never created... It may indeed be safer to lock this import to a specific commit

terratorch/models/prithvi_model_factory.py
CarlosGomes98: Yeah, this PR was definitely too large, so its unfortunate that it steps on some of the work you were doing yourself :( hopefully its not horrible to reconcile these

CarlosGomes98:
> @CarlosGomes98 @romeokienzler General comment. I reviewed the new way to define factories and registries. I believe it simplifies the entire process and introduces clarification for some parts (specially for the registries). Even though it creates some conflicts for two open PRs, I believe it's easier to solve these conflicts when these modifications are already in the main branch. Especially considering that the pretraining and mmseg PRs are not so urgent.

Thanks for the work reviewing this! I know it was a lot :)

pyproject.toml
Joao-L-S-Almeida: I understand. Let's keep as is and see what we can do on the PyPI side. Thanks.

PR #189, opened 2024-10-15
Fix `freeze_decoder`

fmartiescofet:
When freezing the decoder in `PixelWiseModel` and `ScalarOutputModel`, the `encoder` and the `head` were actually being frozen instead of the `decoder` and the `head`.


Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Good. Thanks.

PR #193, opened 2024-10-17
[WIP] Trying to solve conflicts between the EncoderDecoder Factory PR and the Pretraining Task PR

Joao-L-S-Almeida:
Trying to solve conflicts between the EncoderDecoder Factory PR and the remaining patches.


CarlosGomes98:
This approach would make the #186 even bigger :( can we try to either first merge #194 or #186 and then the other one?

Joao-L-S-Almeida:
#194 Is an attempt of running the automated tests, we can ignore it for now. 
My idea was just solving possible conflicts, so if you agree with the modifications carried out here, we could try to merge #186 and after it this PR.

terratorch/tasks/regression_tasks.py
```diff
@@ -115,6 +117,285 @@ def plot(self, val: Tensor | Sequence[Tensor] | None = None, ax=None):
...
+        model_output: ModelOutput = self(x)
+        loss = self.val_loss_handler.compute_loss(model_output, y, self.criterion, self.aux_loss)
+        self.val_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=x.shape[0])
+
+        # Removing the time dimension
+        # TODO This should be less hard-coded
+        # However, it seems that all this part (for all the PixelWise Models) would not work
+        # for models with num_frames > 1.
+        y_hat = self.unpatchify(model_output.output)[:,:,0,...]
+        y = self.unpatchify(y)[:,:,0, ...]
```
CarlosGomes98, line 329: why is this logic now required? if its pre-training specific, It seems like it should be handled by the model. For example, if we are pre-training a CNN, this is not required

terratorch/tasks/regression_tasks.py
```diff
@@ -115,6 +117,285 @@ def plot(self, val: Tensor | Sequence[Tensor] | None = None, ax=None):
     def reset(self) -> None:
         self.metric.reset()
 
+# TODO Maybe we could rewrite this method to inherit from PixelwiseRegressionTask.
```
CarlosGomes98, line 120: I would avoid introducing such a tight coupling. Maybe some of the common logic could be extracted into components which both classes have as attributes

terratorch/models/prithvi_model_factory.py
```diff
@@ -72,27 +77,116 @@ def build_model(
...
+        else: 
+            # By-passing entities not required for pre-training tasks.
+            decoder = None
+            aux_head_kwargs = None 
+            head_kwargs = None 
+            to_be_aux_decoders = None
+
         return self._factory.build_model(task,
                                          backbone,
                                          decoder,
```
CarlosGomes98, line 192: Are these changes required? can it not work by forwarding to the encoder decoder factory, as before?

CarlosGomes98 (commented):
Functionality wise, I think this PR works and makes sense, but I have a few concerns regarding the complexity it would introduce.

If we are adding pretraining, I think it should have its own factory as well as its own model wrapper (a-la pixelwisemodel). In the current approach, I think too much complexity needs to be introduced to these classes.

The encoder_decoder_factory should only know how to combine encoder + decoder. If it also needs to build full models which are actually placed under only the encoder attribute, I believe it has too many responsibilities.

Similarly for the pixel wise model, it should know how to forward through an encoder + decoder structure. If its only a full model for pretraining, a separate, much thinner wrapper should be used, if needed.

I understand this may introduce duplication, but I think it leads to an easier to understand codebase, where each class has fewer responsibilities. To reduce duplication, I would rather opt for identifying common functionalities, packaging them as components, and using composition in these classes where they "have" these components as attributes.

This would also mean we don't need to worry about merging the encoder_decoder_factory pr with this one. encoder_decoder_factory should be able to be merged, and the pretraining should be a completely separate PR that can come after, in my opinion.

terratorch/tasks/regression_tasks.py
Joao-L-S-Almeida: Right.

terratorch/models/prithvi_model_factory.py
Joao-L-S-Almeida: It was an attempt of bypassing the instantiation of a separated decoder considering our legacy infrastructure. With the new EncoderDecoderFactory, maybe we could replace it. 
In any case, it still can work.

Joao-L-S-Almeida:
I don't want to touch this mess. Back to the drawning board.

PR #195, opened 2024-10-18
import custom modules from  dir

CarlosGomes98:
Allow users to add backbones (or other components) without needing to edit the source code. 

All code in the directory `custom_modules` will be imported, so custom modules may be registered

### DIFF
--- a/docs/models.md
+++ b/docs/models.md
@@ -97,3 +97,9 @@ class SMPModelWrapper(Model, nn.Module):
     def freeze_decoder(self):
         pass
 ```
+
+# Custom modules with CLI
+
+Custom modules must be in the import path in order to be registered in the appropriate registries. 
+
+In order to do this without modifying the code when using the CLI, you may place your modules under a `custom_modules` directory. This must be in the directory from which you execute terratorch.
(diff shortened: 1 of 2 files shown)


Joao-L-S-Almeida:
In this way, the user just need to create a directory called `custom_modules` in whataver working directory it is and all the classes and functions defined in the modules inside it will be imported and made available , right ? That seems good. 

I'm not sure about how Lightning is accessing the imported modules, since importlib.import_module returns the module as an object, from which the methods are accessible:
```
In [1]: import importlib

In [2]: arr = importlib.import_module("numpy")

In [3]: arr
Out[3]: <module 'numpy' from '$HOME/.pyenv/versions/3.11.10/envs/terratorch-test/lib/python3.11/site-packages/numpy/__init__.py'>

In [4]: arr.linalg.norm
Out[4]: <function norm at 0x7fbbc4fd9ff0>
```

What if we clone a repository inside `custom_modules`, is importlib supporting an import like `custom_modules.<repo>.<subclass>` ?

Joao-L-S-Almeida (approved):

PR #209, opened 2024-10-31
refactor for gravity wave model integration

romeokienzler:


.gitignore
Joao-L-S-Almeida, line 1: As `config.yaml`is always downloaded from HF, it can be ignored.

Joao-L-S-Almeida (approved):

PR #217, opened 2024-11-01
Custom modules

Joao-L-S-Almeida:
This is a prototype of how we could combine two model pipelines with terratorch. I believe it could be the easiest way to include third-party and custom model in our tests.

### DIFF
--- a/terratorch/tasks/segmentation_tasks.py
+++ b/terratorch/tasks/segmentation_tasks.py
@@ -102,6 +102,7 @@ class SemanticSegmentationTask(BaseTask):
         self._model_module = None
 
         if model_factory:  
+            self.model_factory = MODEL_FACTORY_REGISTRY.build(model_factory)
             self.model_builder = self._build
         else:
             self.model_builder = self._bypass_build
@@ -111,9 +112,6 @@ class SemanticSegmentationTask(BaseTask):
         super().__init__()
 
         self._model_module = model
-
-        if model_factory:
-            self.model_factory = MODEL_FACTORY_REGISTRY.build(model_factory)
 
         
         self.train_loss_handler = LossHandler(self.train_metrics.prefix)
@@ -287,6 +285,7 @@ class SemanticSegmentationTask(BaseTask):
         """
         x = batch["image"]
         y = batch["mask"]
+        print(f"SELF:{self}")
         model_output: ModelOutput = self(x)
         loss = self.val_loss_handler.compute_loss(model_output, y, self.criterion, self.aux_loss)
         self.val_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=x.shape[0])
--- a/terratorch/tasks/segmentation_tasks.py
+++ b/terratorch/tasks/segmentation_tasks.py
@@ -112,7 +112,7 @@ class SemanticSegmentationTask(BaseTask):
         super().__init__()
 
         self._model_module = model
-
+        self.model = model
         
         self.train_loss_handler = LossHandler(self.train_metrics.prefix)
         self.test_loss_handler = LossHandler(self.test_metrics.prefix)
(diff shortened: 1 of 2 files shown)


terratorch/tasks/segmentation_tasks.py
```diff
@@ -98,22 +99,47 @@ def __init__(
...
         self.val_loss_handler = LossHandler(self.val_metrics.prefix)
         self.monitor = f"{self.val_metrics.prefix}loss"
         self.plot_on_val = int(plot_on_val)
 
-    # overwrite early stopping
+    @property
+    def model_module(self):
+        return self._model_module
+
+       # overwrite early stopping
```
romeokienzler, line 127: this comment is in the wrong place?

terratorch/tasks/segmentation_tasks.py
```diff
@@ -98,22 +99,47 @@ def __init__(
...
+            self.model_builder = self._build
+        else:
+            self.model_builder = self._bypass_build
+
+        self._model_module = None 
+
         super().__init__()
+
+        self._model_module = model
+        self.model = model
```
romeokienzler, line 115: What happens if model is not passed as parameter? it is none, correct? does it have any implications?

terratorch/tasks/segmentation_tasks.py
```diff
@@ -98,22 +99,47 @@ def __init__(
...
+
+    def _build(self):
+
+        return self.model_factory.build_model(
             "segmentation", aux_decoders=self.aux_heads, **self.hparams["model_args"]
         )
+
+    def configure_models(self) -> None:
+        self.model: Model = self.model_builder()
+
```
romeokienzler, line 142: So are you sure this works?

Case 1:
You pass a modelfactory, then self.model_builder = self._build and the model is build from the factory

Case 2:
You don't pass a model factory but a model
self.model_builder = self._bypass_build and the one you get passed in from constructor is used

I think the code is correct but hard to understand, I also might wanna add some argument checking, e.g. if model factory is None a model need to be passed

terratorch/tasks/segmentation_tasks.py
Joao-L-S-Almeida: Yep.

terratorch/tasks/segmentation_tasks.py
Joao-L-S-Almeida: I just modified it to guarantee that or a model_factory or a model is provided:
   ```
     if model_factory:  
            self.model_factory = MODEL_FACTORY_REGISTRY.build(model_factory)
            self.model_builder = self._build
        elif model:
            self.model_builder = self._bypass_build
        else:
            raise Exception("Or a model_factory or a torch.nn.Module object must be provided.")
```
Which will allow the selection of the most proper builder method.

File: terratorch/tasks/segmentation_tasks.py (hunk 3 above)
Comment (Joao-L-S-Almeida) [reply terratorch/tasks/segmentation_tasks.py:142]: When no model_factory is passed, it used the object model, so the the method self._bypass_build just return this already instantiated object to be used during the process.

**Comment by romeokienzler** [comment]
@daniszw asking to not merge it now, let's merge mid December :)

File: terratorch/models/backbones/vit_encoder_decoder.py [hunk 4]
```diff
@@ -374,13 +391,29 @@ def _init_weights(self, m):
...
+        return imgs
+
     def patchify(self, imgs: torch.Tensor) -> torch.Tensor:
         """
         imgs: B, C, T, H, W
         x: B, L, D
         """
         p = self.patch_embed.patch_size[0]
         tub = self.patch_embed.tubelet_size
+        imgs = self.pad_images(imgs, patch_size=p)
```
romeokienzler, line 416: @Joao-L-S-Almeida aren't now the images always padded? how do you ensure in case padding==False to skip that function call?

Joao-L-S-Almeida: Now this padding was replaced with that used by @blumenstiel for PrithviViT.
In anyway, I could add an argument to make it optional, but I understand that, when the images are divisible by patch_size, no padding is applied.

romeokienzler:
thanks @Joao-L-S-Almeida 
- [x] merged

PR #228, opened 2024-11-07
Feat: Handle multiple test dataloaders in `SemanticSegmentationTask`

fmartiescofet:
The goal of this PR is to be able to handle multiple test dataloaders in the `SemanticSegmentationTask`.
The purpose of this is because a user might have two test datasets. For example in Sen1Floods11, there are two test datasets: a test dataset which results from splitting all the data except Bolivia flood event into train/val/test and a holdout Bolivia flood event dataset.

This PR should not have any backward compatibility problem as if `test_dataloaders_names` is not passed, the class behaves the same way as before.

To make use of this feature the user should use their own `DataModule` which returns a list of dataloaders in the `test_dataloader` method. `lightning` takes care of evaluate on each of the dataloader sequentially as explained [here](https://lightning.ai/docs/pytorch/stable/data/iterables.html).


terratorch/tasks/segmentation_tasks.py
Joao-L-S-Almeida, line 1: There is backward compatibility since the dataloaders can be reduced to a modules list with a single element.

Joao-L-S-Almeida (approved):

Issue #221, opened 2024-11-04
The GitHub Actions runner is suddenly interrupting the automatic tests. 


Joao-L-S-Almeida:
Actions is interrupting jobs without further information.
![image](https://github.com/user-attachments/assets/e2cf3f7d-e1ef-468e-b728-2a9883cfbacd)

Is it a time limitation ?
Maybe a memory issue ?

romeokienzler:
1. create a parameterize(able) script to consume memory, cpu, execution time
2. slowly increase until failing

Joao-L-S-Almeida:
I tried to increase the timeout (#232), but sometimes it is still faillling, since pytest becames inactive for a long time before starting to run.

romeokienzler:
potentially solved in #232

PR #232, opened 2024-11-11
Limiting the memory used during tests in order to avoid the runner memory overflow. 

Joao-L-S-Almeida:
Maybe it will solve the issue #221 .

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -12,6 +12,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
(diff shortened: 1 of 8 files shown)


Joao-L-S-Almeida:
There is a strict [limitation for the resources of each kind of runner](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories). By explicitly invoking the garbage collector (gc) it seems we have avoided the memory overconsumption for now.

romeokienzler (approved):

PR #233, opened 2024-11-12
Add torchgeo models

paolofraccaro:
Added different models as implemented in torchgeo. Added a related test file and passing all other tests. Only issue is mmseg installation imcompatible with updated version of timm.


terratorch/datasets/__init__.py
Joao-L-S-Almeida, line 1: Is terratorch.datasets.sen1floods11_lat_lon.Sen1Floods11NonGeo going to replace terratorch.datasets.sen1floods11.Sen1Floods11NonGeo ? Or should we keep both ?

Joao-L-S-Almeida:
By running the automatic tests `pytest -s tests/test_encoder_decoder_torchgeo_models.py::test_create_pixelwise_model_dofa`, I see the error:
`FAILED tests/test_encoder_decoder_torchgeo_models.py::test_create_pixelwise_model_dofa[True-IdentityDecoder-segmentation-expected1-dofa_large_patch16_224] - AttributeError: 'DOFA' ob`ject has no attribute 'embed_dim'
In fact, the dofa model here:
terratorch/models/backbones/dofa_vit.py:64: in __init__
    self.out_channels = [self.dofa_model.embed_dim] * len(self.out_indices)
has no attribute called `embed_dim`.
I'm using `timm==0.9.7` and Python 3.12.

Joao-L-S-Almeida:
Should we replace the current torchgeo version from requirements and pyproject.toml by:
`  torchgeo @ git+https://github.com/microsoft/torchgeo `?

paolofraccaro:
@Joao-L-S-Almeida I think it makes sense to get the latest. Or is there a way to point to a specific commit?

Joao-L-S-Almeida:
It's possible to add the hash at the end: `pip install git+https://github.com/microsoft/torchgeo@a5d3037864df8851e2efce6c9ee8134ecf6b1106`

paolofraccaro:
Please also check if there is any other library that requires the same approach.

Joao-L-S-Almeida:
@paolo-fraccaro I tried to use the latest torchgeo but:
`ERROR: Cannot install torchgeo 0.7.0.dev0 (from git+https://github.com/microsoft/torchgeo) and torchgeo 0.7.0.dev0 (from git+https://github.com/microsoft/torchgeo.git) because these package versions have conflicting dependencies.`

terratorch/datasets/__init__.py
romeokienzler: @Joao-L-S-Almeida shall we mark the old one deprecated?

romeokienzler: cc @paolofraccaro ?

romeokienzler (approved):
Thanks @paolofraccaro cc @Joao-L-S-Almeida 

- [x] merged

PR #235, opened 2024-11-14
Feat: Add LORA support in `EncoderDecoderFactory`

fmartiescofet:
This PR adds support for LORA using the [PEFT](https://huggingface.co/docs/peft/index) library.

Some reason behind design choices:
* I used a plain dictionary to pass `peft_config` in the `EncoderDecoderFactory` and delegate the validation to a method so other tools such as benchmark only need to work with standard dictionaries when combining hyperparameters instead of using directly the `TerratorchPEFTConfig` class in the arguments.
* As I only tested LoRA, a warning is shown when a different method is used, but the code should work fine with other methods.
* We only allow to use PEFT with unfrozen `freeze_backbone=False` as the PEFT library already handles freezing/unfreezing each module.

CC: @blumenstiel @smandal94


terratorch/tasks/classification_tasks.py
blumenstiel, line 1: @fmartiescofet Do we need these checks in the tasks? You already included a check in the `EncoderDecoderFactory`. Also these thanks can be used with different factories which might not have peft_config.

blumenstiel: I just saw, that it is a check for `pretrained`, so it is fine here. Maybe change it to `self.hparams.get('peft_config', None)` so that we don't get any other errors.

fmartiescofet: You are right about the different factories which don't have a `peft_config`. I think we can use `get` to avoid a key error. But the check should be done here as the `freeze_backbone` parameter is not passed through to the `EncoderDecoderFactory`.

smandal94:
Reviewed code and trained on Sen4Map for lulc classification. Looks good to me.
Screenshot 2024-11-20 at 06 11 08
version_10554601 is my implementation, version_10576861 is this implementation.

## Issue

During inference/test, I do not need to load pre-trained weight but only the fine-tuned checkpoint weight of the downstream task, which I can pass via:
```
res_test = trainer.test(model=model, ckpt_path=checkpoint_path, dataloaders=data_module)
```

But it's coded to raise ValueError in build_model() which I do not think is necessary for inference.
```
File "/p/project1/geofm4eo/mandal/test/terratorch/terratorch/models/encoder_decoder_factory.py", line 132, in build_model
    raise ValueError(msg)
ValueError: PEFT only works with pretrained backbones
```
I'd suggest raise a warning instead of an error maybe?

terratorch/tasks/classification_tasks.py
romeokienzler: @fmartiescofet is this done?

romeokienzler:
@fmartiescofet thanks for your contrubution, in case u find time would be nice to add some test, merged because the code was non-breaking/low risk

@blumenstiel thanks for your review

PR #237, opened 2024-11-14
[WIP] Generic multimodal dataset and MultiMAE

blumenstiel:
This pr includes:
- A generic multimodal dataset that can process multiple image and non-image modalities (`GenericMultiModalDataModule`)
- MultiMAE backbone (`multimae_base` and `multimae_large`, no pre-trained models)
- The generic dataset comes versions for segmentation, pixel-wise regression, and scalar tasks
- The data config is based on the generic datasets. But instead of passing one `train_data_root`, you can now pass a directory with the modalities as keys.
```
    train_data_root:
      S2L2A: data/sen1floods11/data/data/flood_events/HandLabeled/S2L2AHand
      S1: data/sen1floods11/data/data/flood_events/HandLabeled/S1Hand
      LULC: data/sen1floods11/data/data/flood_events/HandLabeled/LULCHand
```

- Additional parameters are: 
  - `modalities` (list of all modality names) 
  - `image_modalities` (list of image modalities, optional, otherwise considers all modalities as images)
  - `rgb_modality`, `task` (e.g. `segmentation`, pixel-wise `regression`, `scalar` (similar to `classification` etc.))
  - `shared_transforms` (apply the same transformation to all images, default `True`)
  - `chunk_data` (if the data already includes a sample dimension, default `False`)
  - `sample_num_modalities` (int, sample a subset of the modalities per batch, default `None`)
  - `sample_replace` (sample modalities with replacement, default `False`)
  - `concat_bands` (concatenate all images modalities along the band dim which enables running single-modal models ,default `False`)

Example configs: [multimae_sen1floods11.yaml](https://github.com/blumenstiel/terratorch/blob/multimodal/examples/confs/multimae_sen1floods11.yaml) [multimodal_prithvi_sen1floods11.yaml](https://github.com/blumenstiel/terratorch/blob/multimodal/examples/confs/multimodal_prithvi_sen1floods11.yaml)

Open Todos:
- [ ]  Implement `allow_missing_modalities` on a data module level. Idea: Add nan data for missing modalities.
- [x] Update docs
- [ ] Check plotting code for `scalar` tasks
- [ ] Add pre-trained RGB MultiMAE model and weight loading
- [ ] Add tests


romeokienzler:
@daniszw can you please let us know why we should not merge this one now, would it cause conflicts in other PRs?

terratorch/tasks/segmentation_tasks.py
```diff
@@ -240,7 +242,7 @@ def training_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) ->
 
         model_output: ModelOutput = self(x, **rest)
         loss = self.train_loss_handler.compute_loss(model_output, y, self.criterion, self.aux_loss)
-        self.train_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=x.shape[0])
+        self.train_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=y.shape[0])
```
romeokienzler, line 245: @blumenstiel can you please explain why you've changed from x to y?

terratorch/tasks/segmentation_tasks.py
```diff
@@ -278,14 +280,17 @@ def validation_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -
...
-        self.val_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=x.shape[0])
+        self.val_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=y.shape[0])
         y_hat_hard = to_segmentation_prediction(model_output)
         self.val_metrics.update(y_hat_hard, y)
 
         if self._do_plot_samples(batch_idx):
             try:
                 datamodule = self.trainer.datamodule
                 batch["prediction"] = y_hat_hard
+                if isinstance(batch["image"], dict):
```
romeokienzler, line 291: @blumenstiel nooby question, but aren't we loosing flexibility by expecting the image key always to be present? For example in the wxc models it is not present. Or are we saying if there is an image present then we unpack it in case it is packed? or are you using image here some sort of equivalent for 'payload'?

terratorch/tasks/segmentation_tasks.py
```diff
@@ -328,7 +333,7 @@ def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None
         self.test_loss_handler[dataloader_idx].log_loss(
             partial(self.log, add_dataloader_idx=False),  # We don't need the dataloader idx as prefixes are different
             loss_dict=loss,
-            batch_size=x.shape[0],
+            batch_size=y.shape[0],
```
romeokienzler, line 336: @blumenstiel same x y question

romeokienzler (commented):
@blumenstiel @daniszw @PedroConrado I feel this PR is generally safe to merge. I have got 4-5 questions which - if answered - your even make me feel more confident to merge

terratorch/tasks/segmentation_tasks.py
blumenstiel: Because with the new multi-modal dataset x is a dict in order to pass all modalities to the model and does not have `.shape`.

terratorch/tasks/segmentation_tasks.py
```diff
@@ -278,14 +280,17 @@ def validation_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -
         rest = {k: batch[k] for k in other_keys}
         model_output: ModelOutput = self(x, **rest)
         loss = self.val_loss_handler.compute_loss(model_output, y, self.criterion, self.aux_loss)
-        self.val_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=x.shape[0])
+        self.val_loss_handler.log_loss(self.log, loss_dict=loss, batch_size=y.shape[0])
```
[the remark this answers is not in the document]
blumenstiel: Because with the new multi-modal dataset x is a dict in order to pass all modalities to the model and does not have .shape.

terratorch/tasks/segmentation_tasks.py
blumenstiel: It's a segmentation task, so the task definitely expects a key `"image"`. It is also required in the code before and in albumentations.
But we cannot assume that `rgb_modality` is always present in the datamodule. I will change that.

terratorch/tasks/segmentation_tasks.py
blumenstiel: see approve.

terratorch/models/pixel_wise_model.py
```diff
@@ -90,7 +90,11 @@ def _check_for_single_channel_and_squeeze(x):
     def forward(self, x: torch.Tensor, **kwargs) -> ModelOutput:
         """Sequentially pass `x` through model`s encoder, decoder and heads"""
         self.check_input_shape(x)
-        input_size = x.shape[-2:]
+        if isinstance(x, torch.Tensor):
+            input_size = x.shape[-2:]
+        elif isinstance(x, dict):
+            # Multimodal input in passed as dict
+            input_size = list(x.values())[0].shape[-2:]
```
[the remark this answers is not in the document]
blumenstiel: I fix it so that we can always expect an image modality as the first modality from the generic dataset (in case of pixel-wise tasks). I add a functionality that you can also pass the input size in the kwargs to be more flexible for custom datasets.

romeokienzler:
thanks @blumenstiel 
- [x] merged

Issue #236, opened 2024-11-14
Custom module paths


WillTrojak:
Currently, it seems that the location of custom module is both hard coded and implied by the launch location. 

For example, a custom module has to be in a directory called `custom_modules` within the current working directory. 

For example, here:
https://github.com/IBM/terratorch/blob/9912b975ba07a0fcfb9c0d0be3c070e1b9e29cf6/terratorch/cli_tools.py#L59
https://github.com/IBM/terratorch/blob/9912b975ba07a0fcfb9c0d0be3c070e1b9e29cf6/terratorch/cli_tools.py#L390-L397
https://github.com/IBM/terratorch/blob/9912b975ba07a0fcfb9c0d0be3c070e1b9e29cf6/terratorch/registry/custom_registry.py#L6

I would suggest replacing the hard-cored paths, checking for the path in the following order of precedence:
- Command-line flag `--module-path`
- Environment variable `CUSTOM_MODULE_PATH`
- Default to the currently hard-coded path if neither is provided.

WillTrojak:
@romeokienzler @Joao-L-S-Almeida

romeokienzler:
thanks @WillTrojak will put it to the sprint list

PR #251, opened 2024-11-19
Other ways to define custom modules.

Joao-L-S-Almeida:
* custom_modules could be defined via the environment variable TERRATORCH_CUSTOM_MODULE_PATH
* and also via the command line: `--custom_modules_path <path>`

### DIFF
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -118,10 +118,10 @@ def import_custom_modules(custom_modules_path:None | Path | str =None) -> None:
             try:
                 importlib.import_module(module_dir)
                 logger.info(f"Found {custom_modules_path}")
-            except Exception:
-                raise Exception(f"It was not possible to import modules from {custom_modules_path}.")
+            except ImportError:
+                raise ImportError(f"It was not possible to import modules from {custom_modules_path}.")
         else:
-            raise Exception(f"Modules path {custom_modules_path} isn't a directory. Check if you have defined it properly.")
+            raise ValueError(f"Modules path {custom_modules_path} isn't a directory. Check if you have defined it properly.")
     else:
         logger.info("No custom module is being used.")
 
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -389,7 +389,8 @@ class MyLightningCLI(LightningCLI):
 
             custom_modules_path =  self.config.fit.custom_modules_path
         else:
-            custom_modules_path = os.environ.get("CUSTOM_MODULES_DIR_NAME", "custom_modules")
+            default_path = Path(".") / "custom_modules"
+            custom_modules_path = os.environ.get("TERRATORCH_CUSTOM_MODULE_PATH", default_path)
 
         import_custom_modules(custom_modules_path)
 
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -99,20 +99,28 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
     logging.info(f"Saving output to {out_file_name} ...")
     write_tiff(result, os.path.join(out_dir, out_file_name), metadata)
 
-def import_custom_modules(custom_modules_path):
+def import_custom_modules(custom_modules_path:str=None) -> None:
+
     if custom_modules_path:
-        if os.path.isdir(custom_modules_path):
+
+        custom_modules_path = Path(custom_modules_path)
+
+        if custom_modules_path.is_dir():
 
             # Add 'custom_modules' folder to sys.path
-            workdir, module_dir = os.path.split(custom_modules_path)
-
-            if not workdir:
-                workdir = "."
+            workdir = custom_modules_path.parents[0]
+            module_dir = custom_modules_path.name
 
             sys.path.append(workdir)
             logging.info(f"Found {custom_modules_path}")
-
-            importlib.import_module(module_dir)
+            try:
+                importlib.import_module(module_dir)
+            except Exception:
+                raise Exception(f"It was not possible to import modules from {custom_modules_path}.")
+        else:
+            logging.info(f"The modules path {custom_modules_path} not found.")
+    else:
+        logging.info("No custom module is being used.")
 
 class CustomWriter(BasePredictionWriter):
     """Callback class to write geospatial data to file."""
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -56,6 +56,8 @@ from terratorch.tasks import (
     SemanticSegmentationTask,  # noqa: F401
 )
 
+logger = logging.getLogger(__name__)
+
 def flatten(list_of_lists):
     return list(itertools.chain.from_iterable(list_of_lists))
 
@@ -96,7 +98,7 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
     file_name = os.path.basename(input_file_name)
     file_name_no_ext = os.path.splitext(file_name)[0]
     out_file_name = file_name_no_ext + "_pred.tif"
-    logging.info(f"Saving output to {out_file_name} ...")
+    logger.info(f"Saving output to {out_file_name} ...")
     write_tiff(result, os.path.join(out_dir, out_file_name), metadata)
 
 def import_custom_modules(custom_modules_path:str=None) -> None:
@@ -112,15 +114,16 @@ def import_custom_modules(custom_modules_path:str=None) -> None:
             module_dir = custom_modules_path.name
 
             sys.path.append(workdir)
-            logging.info(f"Found {custom_modules_path}")
+
             try:
                 importlib.import_module(module_dir)
+                logger.info(f"Found {custom_modules_path}")
             except Exception:
                 raise Exception(f"It was not possible to import modules from {custom_modules_path}.")
         else:
-            logging.info(f"The modules path {custom_modules_path} not found.")
+            logger.info(f"The modules path {custom_modules_path} not found.")
     else:
-        logging.info("No custom module is being used.")
+        logger.info("No custom module is being used.")
 
 class CustomWriter(BasePredictionWriter):
     """Callback class to write geospatial data to file."""
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -3,7 +3,7 @@
 import importlib.util
 import itertools
 import json
-import logging  # noqa: I001
+import logging
 import os
 import shutil
 import sys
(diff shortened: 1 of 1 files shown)


Joao-L-S-Almeida:
Trying to solve https://github.com/IBM/terratorch/issues/236.
Is it what you expect to have @WillTrojak.

terratorch/cli_tools.py
```diff
@@ -101,6 +99,20 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
     logging.info(f"Saving output to {out_file_name} ...")
     write_tiff(result, os.path.join(out_dir, out_file_name), metadata)
 
+def import_custom_modules(custom_modules_path):
+    if custom_modules_path:
```
WillTrojak, line 103: `custom_modules_path` should probably have a default of `None`

terratorch/cli_tools.py
```diff
@@ -101,6 +99,20 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
...
+        if os.path.isdir(custom_modules_path):
```
WillTrojak, line 104: Then here it might be better to do `custom_modules_path = Path(custom_modules_path)` before then doing the if statement. pathlib's Path class is a lot more portable than calls to `os`

terratorch/cli_tools.py
```diff
@@ -101,6 +99,20 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
...
+
+            # Add 'custom_modules' folder to sys.path
+            workdir, module_dir = os.path.split(custom_modules_path)
+
+            if not workdir:
```
WillTrojak, line 109: I think this line assumes that a path that does not have a `workdir` must therefore be in the current directory. For example, if `custom_modules_path = "/module"`. You would then erroneously add `.` when you should add `/`. I think if you look at the method in the Path class, you should be able to find something better than manually splitting the path.

terratorch/cli_tools.py
WillTrojak, line 104: I think you should raise an exception is the `custom_modules_path` is not None and doesn't exist.

terratorch/cli_tools.py
```diff
@@ -101,6 +99,20 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
...
+                workdir = "."
+
+            sys.path.append(workdir)
+            logging.info(f"Found {custom_modules_path}")
+
+            importlib.import_module(module_dir)
```
WillTrojak, line 115: I think you need a try-except here because what if you try to import the custom module, but it fails? If you catch the exception, you can probably give the user a more helpful error than whatever importlib does.

terratorch/cli_tools.py
```diff
@@ -352,13 +366,21 @@ def instantiate_classes(self) -> None:
...
+
         if hasattr(config, "out_dtype"):
             self.trainer.out_dtype = config.out_dtype
 
         if hasattr(config, "deploy_config_file"):
             self.trainer.deploy_config = config.deploy_config_file
 
+        # Custom modules path
+        if hasattr(self.config.fit, "custom_modules_path"):
+
```
WillTrojak, line 378: Whitespace

terratorch/cli_tools.py
```diff
@@ -352,13 +366,21 @@ def instantiate_classes(self) -> None:
...
+            custom_modules_path =  self.config.fit.custom_modules_path
+        else:
+            custom_modules_path = os.environ.get("CUSTOM_MODULES_DIR_NAME", "custom_modules")
```
WillTrojak, line 381: maybe set the default to be an actual relative path. Given the point I raised above.
So I would do something like:
```
default_path = Path(".") / "custom_modules"
custom_modules_path = os.environ.get("TERRATORCH_CUSTOM_MODULE_PATH", default_path)
```

terratorch/cli_tools.py
```diff
@@ -101,6 +99,20 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
     logging.info(f"Saving output to {out_file_name} ...")
     write_tiff(result, os.path.join(out_dir, out_file_name), metadata)
 
+def import_custom_modules(custom_modules_path):
+    if custom_modules_path:
+        if os.path.isdir(custom_modules_path):
+
+            # Add 'custom_modules' folder to sys.path
+            workdir, module_dir = os.path.split(custom_modules_path)
```
WillTrojak, line 107: You may want a way of handling `:` separated paths. Ie if someone sets an environment variable with multiple paths, it is fairly standard to separate these with a `:`. Just a thought, otherwise you may want to throw an error for that case.

WillTrojak (changes requested):
This is boardly what I had in mind, but I think there are some points that need addressing.

terratorch/cli_tools.py
Joao-L-S-Almeida: I agree this is a topic to keep in mind.

Joao-L-S-Almeida:
Thanks you for the review, @WillTrojak.

terratorch/cli_tools.py
```diff
@@ -98,9 +98,32 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
     file_name = os.path.basename(input_file_name)
     file_name_no_ext = os.path.splitext(file_name)[0]
     out_file_name = file_name_no_ext + "_pred.tif"
-    logging.info(f"Saving output to {out_file_name} ...")
+    logger.info(f"Saving output to {out_file_name} ...")
     write_tiff(result, os.path.join(out_dir, out_file_name), metadata)
 
+def import_custom_modules(custom_modules_path:str=None) -> None:
```
WillTrojak, line 104: The type hint I think should be `custom_modules_path: None | Path | str=None`

terratorch/cli_tools.py
```diff
@@ -98,9 +98,32 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
...
+
+            sys.path.append(workdir)
+
+            try:
+                importlib.import_module(module_dir)
+                logger.info(f"Found {custom_modules_path}")
+            except Exception:
+                raise Exception(f"It was not possible to import modules from {custom_modules_path}.")
+        else:
+            logger.info(f"The modules path {custom_modules_path} not found.")
```
WillTrojak, line 124: I don't think we want to just log this, I think we want to raise an error. For example:
`raise ValueError(f"Custom module path '{custom_modules_path}' not found")`

terratorch/cli_tools.py
```diff
@@ -98,9 +98,32 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
...
+            # Add 'custom_modules' folder to sys.path
+            workdir = custom_modules_path.parents[0]
+            module_dir = custom_modules_path.name
+
+            sys.path.append(workdir)
+
+            try:
+                importlib.import_module(module_dir)
+                logger.info(f"Found {custom_modules_path}")
+            except Exception:
```
WillTrojak, line 121: Does this even work? I think you need to catch a specific exception.

terratorch/cli_tools.py
```diff
@@ -352,13 +377,22 @@ def instantiate_classes(self) -> None:
...
+
         if hasattr(config, "out_dtype"):
             self.trainer.out_dtype = config.out_dtype
 
         if hasattr(config, "deploy_config_file"):
             self.trainer.deploy_config = config.deploy_config_file
 
+        # Custom modules path
+        if hasattr(self.config.fit, "custom_modules_path"):
+
```
WillTrojak, line 389: Remove new line or use a linter like ruff to do this automatically.

terratorch/cli_tools.py
```diff
@@ -344,6 +368,7 @@ def add_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
         parser.link_arguments("ModelCheckpoint.dirpath", "StateDictModelCheckpoint.dirpath")
 
     def instantiate_classes(self) -> None:
+
```
WillTrojak, line 371: Remove newline or use a linter like ruff to do this automatically.

WillTrojak (changes requested):
It's looking better but I think some work is needed around the exception handling.

terratorch/cli_tools.py
```diff
@@ -98,9 +98,32 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
...
+            sys.path.append(workdir)
+
+            try:
+                importlib.import_module(module_dir)
+                logger.info(f"Found {custom_modules_path}")
+            except Exception:
+                raise Exception(f"It was not possible to import modules from {custom_modules_path}.")
+        else:
+            raise Exception(f"Modules path {custom_modules_path} isn't a directory. Check if you have defined it properly.")
```
WillTrojak, line 124: `Exception` isn't very useful as it's really just a base exception class and it's considered bad practise to raise that. Could you raise a `ValueError`? The issue is that is later, if someone wants to use a `try except` on this function they are forced to catch every exception which isn't very useful.

terratorch/cli_tools.py
```diff
@@ -98,9 +98,32 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
...
+            # Add 'custom_modules' folder to sys.path
+            workdir = custom_modules_path.parents[0]
+            module_dir = custom_modules_path.name
+
+            sys.path.append(workdir)
+
+            try:
+                importlib.import_module(module_dir)
+                logger.info(f"Found {custom_modules_path}")
+            except Exception:
```
WillTrojak, line 121: Could you find out what specific exception you would get in this case and catch that? Because at the moment, you'll catch all exceptions and then give it the new message below, which might not be correct. Also, once caught, use a `from` statement.

WillTrojak (changes requested):
I think you need to be specific about which exception you are catching and raising

terratorch/cli_tools.py
Joao-L-S-Almeida: What about these choices. 

```
    if custom_modules_path:

        custom_modules_path = Path(custom_modules_path)

        if custom_modules_path.is_dir():

            # Add 'custom_modules' folder to sys.path
            workdir = custom_modules_path.parents[0]
            module_dir = custom_modules_path.name

            sys.path.append(workdir)

            try:
                importlib.import_module(module_dir)
                logger.info(f"Found {custom_modules_path}")
            except ImportError:
                raise ImportError(f"It was not possible to import modules from {custom_modules_path}.")
        else:
            raise ValueError(f"Modules path {custom_modules_path} isn't a directory. Check if you have defined it properly.")
    else:
        logger.info("No custom module is being used.")

```

WillTrojak (approved):
Looks great, thanks for sorting this.

PR #259, opened 2024-11-26
Dependabot target is now maintenance

Joao-L-S-Almeida:


romeokienzler:
@Joao-L-S-Almeida what was the intention? Is 'maintenance'==active release which is installed and used?

I think we could define:

maintenance==latest stable version (and just move the maintenance branch forward on each release)

Question, shouldn't we keep depenabot on the main branch as well to spot potential issues early on?

romeokienzler:
@Joao-L-S-Almeida will merge for now

PR #260, opened 2024-11-26
[WIP] Maintenance

Joao-L-S-Almeida:
Periodical evaluation of dependencies being updated.


terratorch/cli_tools.py
```diff
@@ -387,15 +421,6 @@ def build_lightning_cli(
                 stacklevel=1,
             )
 
-    # import any custom modules
```
romeokienzler, line 390: why is this removed?

romeokienzler:
@Joao-L-S-Almeida custom modules related code has been removed in a maintenance PR, was this on purpose?

PR #266, opened 2024-11-28
Updated prithvi mae code

blumenstiel:


Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
All the modifications are good to me and I don't see any conflct with our current status, since some legacy models will be kept in order to enforce back-compatibility (as TemporalViTEncoder). However, it's literally impossible to say if seomthing will break when we try to merge the other PRs still open. In anyway, let's don't have too many irons in the fire for now and just finish it. 
I'll merge.

blumenstiel:
@Joao-L-S-Almeida Please wait until @daniszw and I have updated the model code. We have some further modifications.

PR #268, opened 2024-12-02
Fix `state_dict` redefinition

fmartiescofet:
When filtering the checkpoint weights, `state_dict` was being redefined while looping through it. This did not cause any issue or bug as it was never referenced again inside the loop and python "locks" the items it when starting iterating over it but could give future problems if the code was changed.

CC: @blumenstiel


blumenstiel (approved):
Looks good, thanks for catching this!

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Good. Thank you @fmartiescofet @blumenstiel

PR #275, opened 2024-12-03
Init prithvi models from hf configs

blumenstiel:
Changed from hard coded configs to hf configs to be better aligned with the model weights and timms approach.

This change is also required to track download counts in hf.


terratorch/models/backbones/prithvi_mae.py
Joao-L-S-Almeida, line 1: I understand the need of controlling the parameter eps, but maybe it would be better to input it as an argument in the future. The default value for eps is 1e-05, so the reason for using 1e-06 isn't clear.

terratorch/models/backbones/prithvi_vit.py
Joao-L-S-Almeida, line 1: All the `create_prithvi_vit*`  functions were replaced with a single one. Right.

Joao-L-S-Almeida (approved):

PR #281, opened 2024-12-04
Solving issue with test stage

Joao-L-S-Almeida:


blumenstiel (changes requested):
You need to use
```
if hasattr(self.config, "fit") and  hasattr(self.config.fit, "custom_modules_path"):
```

Otherwise, it does not go into `else.

Joao-L-S-Almeida:
Done.

blumenstiel:
We cannot merge it this pr because of the force push.

PR #299, opened 2024-12-05
Improve/custom 

Joao-L-S-Almeida:
This PR will replace  #285.


terratorch/tasks/classification_tasks.py
```diff
@@ -96,21 +97,36 @@ def __init__(
...
     def configure_callbacks(self) -> list[Callback]:
         return []
 
     def configure_models(self) -> None:
+
+        if not hasattr(self, "model_factory"):
+            # Custom model is provided
+            if self.hparams["freeze_backbone"] or self.hparams["freeze_decoder"]:
+                logger.warning("freeze_backbone and freeze_decoder are ignored if a custom model is provided.")
+            return
```
romeokienzler, line 124: @Joao-L-S-Almeida doesn't it mean configure_models will always return if model_factory is not set?

terratorch/tasks/regression_tasks.py
```diff
@@ -198,9 +206,17 @@ def configure_callbacks(self) -> list[Callback]:
         return []
 
     def configure_models(self) -> None:
+
+        if not hasattr(self, "model_factory"):
+            # Custom model is provided
+            if self.hparams["freeze_backbone"] or self.hparams["freeze_decoder"]:
+                logger.warning("freeze_backbone and freeze_decoder are ignored if a custom model is provided.")
+            return
```
romeokienzler, line 214: @Joao-L-S-Almeida same problem as above...

terratorch/tasks/regression_tasks.py
```diff
@@ -401,5 +421,5 @@ def model_forward(x):
         if self.tiled_inference_parameters:
             y_hat: Tensor = tiled_inference(model_forward, x, 1, self.tiled_inference_parameters)
         else:
-            y_hat: Tensor = self(x).output
+            y_hat: Tensor = self(x, **rest).output
```
romeokienzler, line 424: @Joao-L-S-Almeida can you please add a comment in the code what **rest is and why it makes sense? Will make maintenance easier

terratorch/tasks/segmentation_tasks.py
```diff
@@ -106,8 +108,21 @@ def __init__(
         self.tiled_inference_parameters = tiled_inference_parameters
         self.aux_loss = aux_loss
         self.aux_heads = aux_heads
-        self.model_factory = MODEL_FACTORY_REGISTRY.build(model_factory)
+
+        if model is not None and model_factory is not None:
+            logger.warning("A model_factory and a model was provided. The model_factory is ignored.")
+        assert model is not None or model_factory is not None, \
```
romeokienzler, line 114: can we raise an error instead? assertions can be switched off

terratorch/tasks/segmentation_tasks.py
```diff
@@ -121,9 +136,16 @@ def configure_callbacks(self) -> list[Callback]:
         return []
 
     def configure_models(self) -> None:
+        if not hasattr(self, "model_factory"):
+            # Custom model is provided
+            if self.hparams["freeze_backbone"] or self.hparams["freeze_decoder"]:
+                logger.warning("freeze_backbone and freeze_decoder are ignored if a custom model is provided.")
+            return
```
romeokienzler, line 143: @Joao-L-S-Almeida see same problem as above

terratorch/tasks/classification_tasks.py
blumenstiel: @romeokienzler That's how it should work. If model_factory is not provided, a custom model is passed. Therefore, the rest of the function should be skipped.

terratorch/tasks/segmentation_tasks.py
blumenstiel: The code is okay.

blumenstiel (approved):

terratorch/tasks/regression_tasks.py
blumenstiel: Same structure as in the other steps. It's additional keys from the dataset (e.g. location or timestamp).

PR #304, opened 2024-12-06
Prithvi v1 renaming and config fix

blumenstiel:
Changes:
- Renamed prithvi v1 to `prithvi_eo_v1_100`. The old names are still working
- Only download config from hf for pre-trained models, otherwise use default ones (fix wrong hf counting due to terratorch tests)
- Modified logging messages


terratorch/models/backbones/prithvi_mae.py
Joao-L-S-Almeida, line 1: It was already modified in another way for #299. It will require further merging.

Joao-L-S-Almeida (approved):

PR #311, opened 2024-12-09
Fixing

Joao-L-S-Almeida:
Fixing errors associated with #307


blumenstiel (approved):
Thanks for fixing it!

PR #318, opened 2024-12-11
Fix timm pretrained error

blumenstiel:
Timm expects the pretrained configs to be named `default_cfgs`. Renaming it resulted in not landing pre-trained weights at all.


blumenstiel:
Thanks @fmartiescofet for noticing the error!

Joao-L-S-Almeida:
Good. Thanks.

Joao-L-S-Almeida (approved):

PR #328, opened 2024-12-16
201modularize wxc

romeokienzler:


terratorch/models/pincers/unet_pincer.py
Joao-L-S-Almeida, line 1: The U-Net-like desgine seems very clear and consistent. My only question is if the choices made for activations (ReLU), kernel size and stride should be optionally chosen by the user or we can always go forward with this default choice.

terratorch/models/wxc_model_factory.py
Joao-L-S-Almeida, line 1: I understand that the previous options ("gravitywave" and else) were kept unchanged, so is the option "prithviwxc" encompassing all the finetuning tasks which will modify both encoder and decoder (wrapping the backbone) ? Is it going to replace the others options ?

Joao-L-S-Almeida (approved):

terratorch/models/pincers/unet_pincer.py
romeokienzler: Thanks, this is a good point, I think we need to provide a way to let the user choose. I'm about to incorporate the wxc downscaling task in a modular way as well, so I assume I need to add even more flexibility on the unet

PR #329, opened 2024-12-18
Feat: Implement option to have multiple learning rates

fmartiescofet:
This PR implements the option `lr_overrides` in all the tasks which allow the user to use custom learning rates for different parameters (e.g. it may want to have a lower lr in the encoder vs the decoder).

CC: @blumenstiel @paolofraccaro


terratorch/tasks/segmentation_tasks.py
Joao-L-S-Almeida, line 1: As I could understand, we can provide a dictionary informing which learning rate to use to update each parameter in the model, right ? Is it meaning we need to (optionally) pass a dict like: `{"encoder.layer_1.weight" 1e-3, ...}` ? Could you provide and example (yaml) and a unit test for it ?

terratorch/tasks/base_task.py
Joao-L-S-Almeida, line 1: `parameters` is a list containing multiple dictionaries `{"params": parameters}` which will be send to the optimizer, right ? When these dictionaries contain an extra key "lr", the optimizer replaces the global lr with this one for that specific parameters, right ?

Joao-L-S-Almeida: The content of `self.parameters()` is exactly equal to `self.model.named_parameters()` ?

fmartiescofet: To the first question yes, if the parameter group has a different learning rate this will be used instead of the defaul, if the group does not have a custom lr, it will use the default, see [this](https://pytorch.org/docs/stable/optim.html#per-parameter-options) explanation in pytorch docs.

For the second in our case I believe it does not matter as the task itself does not contain any parameters, either way, I replaced it so we get the parameters from the task instead if the model.

terratorch/tasks/segmentation_tasks.py
fmartiescofet: Yes, the idea for this is that the user can provide a dictionary where the key is a string which should be a substring of all the parameters they want to have that lr, e.g. `"encoder"`, then the value is going to be for all the parameter containing `"encoder"` in their name.
Obviously if they want to become more selective they could do it like `{"encoder.layer_1.weight" 1e-3, ...}`...

I thought of also implementing it in regex strings, but it may be harder for the users to know how to deal with them properly so I think the substring method is easier.

I added an example and to the task creation tests.

Joao-L-S-Almeida (approved):

PR #330, opened 2024-12-18
Feat: Implement multiple test dataloaders in all tasks

fmartiescofet:
To avoid having to override `on_test_epoch_end` on `SemanticSegmentationTask` due to #228. This PR implements it also on `RegressionTask` and `ClassificationTask`.


fmartiescofet:
I can't reproduce the failing tests on local and as far as I can tell they are not related to the changes. Any ideas @Joao-L-S-Almeida @romeokienzler ?

Joao-L-S-Almeida (approved):

Foxigod:
Just a note, this caused backwards compatibility issues for me.

The metrics are (and have always been) hard-coded in the `configure_metrics()` function https://github.com/fmartiescofet/terratorch/blob/e641e7cf63cc6fa059d76163187f931771d6cd1a/terratorch/tasks/classification_tasks.py#L161 and hardly configurable via inputs. For this reason I had a custom class that inherited from the `ClassificationTask` class: https://github.com/fmartiescofet/terratorch/blob/e641e7cf63cc6fa059d76163187f931771d6cd1a/terratorch/tasks/classification_tasks.py#L28 and modified the `configure_metrics()` function to include my metrics of interest.

The issue then arose because this change no longer handles the `self.test_metrics` when it is provided as an instance of `torchmetrics.MetricCollection`, but requires it to be an _iterable_/`nn.ModuleList` of e.g. `torchmetrics.MetricCollection`.

\
After identifying what caused the issue, I was able to modify my custom local version of the classification task to create the `self.test_metrics` variable in a valid way. I suppose I don't expect many others to face the same issue, but nevertheless wanted to mention this.

PR #350, opened 2025-01-06
Fix/select weights

Joao-L-S-Almeida:


Joao-L-S-Almeida:
Unless I expect they will not be cancelled. However, they are still failling for some of the new torchgeo-related tests. I'll try to fix it.

PR #351, opened 2025-01-07
increasing timeout for unit tests and fixing issues with tests

Joao-L-S-Almeida:
Is 30 minutes a suffiicent time ?
Some tests were modified in order to solve issues during the execution.

### DIFF
--- a/terratorch/models/backbones/select_patch_embed_weights.py
+++ b/terratorch/models/backbones/select_patch_embed_weights.py
@@ -48,16 +48,19 @@ def select_patch_embed_weights(
             _possible_keys_for_proj_weight = {custom_proj_key}
 
         patch_embed_proj_weight_key = state_dict.keys() & _possible_keys_for_proj_weight if (type(state_dict) in [collections.OrderedDict, dict]) else state_dict().keys() & _possible_keys_for_proj_weight
+
         if len(patch_embed_proj_weight_key) == 0:
             msg = "Could not find key for patch embed weight"
             raise Exception(msg)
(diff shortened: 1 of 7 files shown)


blumenstiel (approved):
Let's check if the tests run successful in 30 min before merging.

Joao-L-S-Almeida:
The tests took ~ 1100 s, so it seem the problem are still some new tests that are failing. I'll try to fix them here.

terratorch/models/backbones/select_patch_embed_weights.py
```diff
@@ -78,26 +83,4 @@ def select_patch_embed_weights(
...
-        warnings.warn(
-            f"Incompatible shapes between patch embedding of model {temp_weight.shape} and\
-            of checkpoint {patch_embed_weight.shape}",
-            category=UserWarning,
-            stacklevel=1,
-        )
-
-    state_dict[patch_embed_proj_weight_key] = temp_weight
-    return state_dict
+        return state_dict
```
romeokienzler, line 86: @Joao-L-S-Almeida why is it safe to remove that code block?

Joao-L-S-Almeida: In fact, is not being removed, I erroneously replicated this block during  a merging. See [here](https://github.com/IBM/terratorch/blob/2614193f7add54c4e89cfb68d03e783952c5d646/terratorch/models/backbones/select_patch_embed_weights.py#L65) until line 86, this block is still maintained.

PR #357, opened 2025-01-09
Feat: Implement Terratorch UNet decoder

fmartiescofet:
`smp_UNet` decoder does not use the first embedding as they assume it has the same resolution as the original image, see [here](https://github.com/qubvel-org/segmentation_models.pytorch/issues/112#issuecomment-565712493), that's usually true for `UNet` encoders but not in the case we use a plain ViT and we upsample the output embeddings. For this reason I created a `UNetDecoder` in `terratorch` which is built around `smp_Unet` decoder but avoids skipping this first embedding. This is more intuitive as it's how other decoders work (e.g. UperNet).

CC: @blumenstiel


blumenstiel:
Thanks @fmartiescofet, looks good to me!

@Joao-L-S-Almeida @romeokienzler

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Hi @blumenstiel and @fmartiescofet I think all the additions are ok. Let's approve it.

PR #377, opened 2025-01-21
Refactor prithvi

blumenstiel:
Changes:
- Initaitze prithvi backbone via the TerraTroch backbone registry instead of timm.
- Simplified registration code
- Updated test to work with updated prithvi code
- Added custom errors for users requesting the old model names (should be removed after one terratorch version update).
- Fixed `jsonargparse` version due to CLI failing
- Added interpolated pos embedding instead of extrapolation
- Updated warnings
- Updated yaml examples
- Added `drop_path` to Prithvi


examples/confs/burn_scars.yaml
Joao-L-S-Almeida, line 1: Why the complete path is required ? Is it conflicting with another module. I found no equivalent in torchgeo. 
Same for the occurrences seen in the other yaml files.

requirements/required.txt
Joao-L-S-Almeida, line 1: Wouldn't be better to use `jsonargparse<=4.35.0` as seen for the `pyproject.toml`.

terratorch/models/backbones/dofa_vit.py
Joao-L-S-Almeida, line 1: Is it related to the recent changes in torchgeo ? Should we update the current version in the installation files ?

Joao-L-S-Almeida:
Review in progress (RIP).

terratorch/models/backbones/prithvi_vit.py
blumenstiel, line 1: We should remove the timm module from weight paths to support old model weights trained with previous terratorch versions.

examples/confs/burn_scars.yaml
blumenstiel: No but we agreed with @romeokienzler and @daniszw that we should specify the datamodule paths as a best practice. That's why I changed it in the examples. So that people know if it is comming from torchgeo or terratorch or a custom one

terratorch/models/backbones/prithvi_vit.py
blumenstiel: Done

romeokienzler (approved):

PR #380, opened 2025-01-23
add cli support for wxc gravity wave

romeokienzler:


terratorch/cli_tools.py
Joao-L-S-Almeida, line 1: Maybe it would be better avoiding the print statement in line 165.

Joao-L-S-Almeida (approved):

PR #382, opened 2025-01-23
Incorporating LoRA for GFMs into Terratorch 

Michal-Muszynski:
The goal is to incorporate LoRA into Terratorch.


fmartiescofet:
Hi @Michal-Muszynski, LoRA was already implemented in #235 using the [peft](https://huggingface.co/docs/peft/index) library. You can see an example on how to use it [here](https://github.com/IBM/terratorch/blob/main/examples/confs/sen1floods11_vit_peft.yaml). Let me know if you have any issues.

romeokienzler:
Oh thanks @fmartiescofet and sorry @Michal-Muszynski for the false information

PR #388, opened 2025-01-28
Replacing hash-version with stable torchgeo==0.6.2 and segmentation-models-pytorch==0.4.0

Joao-L-S-Almeida:
These releases already encompass the features required by the hash-versions currently used.


Joao-L-S-Almeida:
No more necessary.

PR #391, opened 2025-01-30
weights_only=True for all the occurences of torch.load

Joao-L-S-Almeida:
It will avoid PyTorch warnings and make the loading process more secure, since it won't load possible malicious code hidden in the checkpoints.


romeokienzler:
@romeokienzler test with a local checkpoint, create test ...

romeokienzler (dismissed):
lgtm

Joao-L-S-Almeida:
I think it's still not approved.

romeokienzler (approved):
lgtm, all test passing on the branch

- [x] merged

PR #397, opened 2025-02-03
fix padding

Joao-L-S-Almeida:


blumenstiel:
@Joao-L-S-Almeida Let's add some test for multi-temporal padding before merging.

romeokienzler:
checking with @blumenstiel if we can close

romeokienzler:
spoken to @blumenstiel, we need to test this one and then merge

romeokienzler (approved):
lgtm

PR #412, opened 2025-02-06
fix-376

### LINKED ISSUE #376: Sen1floods11 bug
**Describe the issue**
The sen1floods11 is pointing not to the right dataset but [`sen1floods11_lat_lon.py`](https://github.com/IBM/terratorch/blob/main/terratorch/datasets/sen1floods11_lat_lon.py) (this is imported in the [`__init__.py`](https://github.com/IBM/terratorch/blob/main/terratorch/datasets/__init__.py) of the datasets).

**To Reproduce (optional, but appreciated)**
Any training using the sen1floods11 dataset.

**To Solve (optional, but appreciated)**
Use the dataset coded in [`sen1floods11.py`](https://github.com/IBM/terratorch/blob/main/terratorch/datasets/sen1floods11.py). Test that this is working with lat/lon and temporal metadata too.

PedroConrado:
fixes #376 issue

### DIFF
--- a/terratorch/datasets/__init__.py
+++ b/terratorch/datasets/__init__.py
@@ -37,7 +37,7 @@ from terratorch.datasets.pastis import PASTIS
 
 # GenericNonGeoRegressionDataset,
 
-from terratorch.datasets.sen1floods11_lat_lon import Sen1Floods11NonGeo
+from terratorch.datasets.sen1floods11 import Sen1Floods11NonGeo
 from terratorch.datasets.utils import HLSBands, OpticalBands, SARBands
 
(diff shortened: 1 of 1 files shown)


Joao-L-S-Almeida:
The two datasets look very similar, but for some minor differences. Are you sure they are totally equivalent ?

PedroConrado:
They are equivalent, but the sen1floods11_lat_lon.py dataset is broken. The sen1floods11.py dataset has been used in the DataModule from the beginning. As I mentioned in the issue, I’m not sure how two equivalent datasets were introduced, but sen1floods11.py is the one that is currently functional and has always been used in the DataModule.

Joao-L-S-Almeida (approved):

PR #415, opened 2025-02-07
Trying to solve issue with replicated input arguments during model (ViT) instantiation/loading

Joao-L-S-Almeida:


romeokienzler (approved):
lgtm

PR #417, opened 2025-02-11
Adjusting the weights keys when necessary

Joao-L-S-Almeida:
It tries to circumvent issues when restoring from checkpoints whose the weights have different names (usually with the addition of a prefix).
This PR tries to solve the question for now, but for the first stable release, we need to find a way to avoid this strange behavior.
I already merged the branch multiclass from PR https://github.com/IBM/terratorch/pull/393 in order to extend the tests.


Joao-L-S-Almeida:
I think it's still not approved.

romeokienzler (approved):
thanks @Joao-L-S-Almeida 

lgtm

- [x] merged

PR #445, opened 2025-02-20
Invoking gc for all these tests

Joao-L-S-Almeida:
It should avoid memory overflow in the GH runners.


PedroConrado (approved):
Thanks @Joao-L-S-Almeida. I forgot the gc.

Joao-L-S-Almeida:
In fact, I don't know if this is essential, but as another PR was suddently interrupted, so I suspect memory overflow could be the reason for it.

PR #451, opened 2025-02-21
Disabling stackability test when requested

Joao-L-S-Almeida:


paolofraccaro:
any reason why sometimes it is false and sometimes is true? implementation looks good. Did we add a test?

Joao-L-S-Almeida:
@paolofraccaro Just a typo. Fixed in the last commit. I tested it manually using:
 `terratorch fit -c manufactured-finetune_prithvi_eo_v2_300.yaml
`

romeokienzler (approved):
thanks @Joao-L-S-Almeida 

lgtm

PR #453, opened 2025-02-21
UNet in the register

Joao-L-S-Almeida:
UNet and ASPPHeads weren't visible in the register.


Joao-L-S-Almeida:
#457 already does it.

PR #454, opened 2025-02-24
ModelCheckpoint must be defined in the config dict, not during the parsing.

Joao-L-S-Almeida:
It should avoid the issue:
```
    raise MisconfigurationException(
lightning.fabric.utilities.exceptions.MisconfigurationException: Trainer was configured with `enable_checkpointing=False` but found `ModelCheckpoint` in callbacks list.

```

### DIFF
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -110,7 +110,7 @@ def add_default_checkpointing_config(config):
 
     if enable_checkpointing:
         if not there_is_checkpointing:
-            print("Enabling ModelCheckpoint since the user defined enable_checkpointing=True.")
+            logger.info("Enabling ModelCheckpoint since the user defined enable_checkpointing=True.")
 
             config["ModelCheckpoint"] = StateDictAwareModelCheckpoint
             config["ModelCheckpoint.filename"] = "{epoch}"
@@ -120,7 +120,7 @@ def add_default_checkpointing_config(config):
             config["StateDictModelCheckpoint.save_weights_only"] = True
             config["StateDictModelCheckpoint.monitor"] = "val/loss"
         else:
-            print("No extra checkpoint config will be added, since the user already defined it in the callbacks.")
+            logger.info("No extra checkpoint config will be added, since the user already defined it in the callbacks.")
 
     return config 
 
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -393,22 +393,6 @@ class MyLightningCLI(LightningCLI):
         parser.add_argument("--deploy_config_file", type=bool, default=True)
         parser.add_argument("--custom_modules_path", type=str, default=None)
 
-        # parser.set_defaults({"trainer.enable_checkpointing": False})
-        """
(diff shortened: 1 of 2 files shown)


terratorch/cli_tools.py
```diff
@@ -96,6 +96,23 @@ def write_tiff(img_wrt, filename, metadata):
             dest.write(img_wrt[i, :, :], i + 1)
     return filename
 
+def add_default_checkpointing_config(config):
+
+    subcommand = config["subcommand"]
+    enable_checkpointing = config[subcommand + ".trainer.enable_checkpointing"]
+
+    if enable_checkpointing:
+        print("Enabling ModelCheckpoint since the user defined enable_checkpointing=True.")
```
blumenstiel, line 105: @Joao-L-S-Almeida Can we avoild prints and use loggers? It is better practice

terratorch/cli_tools.py
```diff
@@ -96,6 +96,23 @@ def write_tiff(img_wrt, filename, metadata):
...
+
+        config["ModelCheckpoint"] = StateDictAwareModelCheckpoint
```
blumenstiel, line 107: I know this is the way Carlos  implemented the checkpoints.

But in lightning, checkpoints are part of the callbacks: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.ModelCheckpoint.html

I am not sure why follow a different pattern, it might be easier for users to overwrite the defaults if we follow the lightning approach (beeing in line with their docs). Otherwise, we need to describe it well in the docs. 
@CarlosGomes98 any arguments why the model checkpoint is not implemented as a callback?

blumenstiel: Also, the setting by the user is just overwritten with this code. So there is no possibility that a user can change this setting.

blumenstiel (dismissed):

terratorch/cli_tools.py
Joao-L-S-Almeida: I agree. This PR is just a way to circumvent the inconvenient fact that model checkpointing is technically obligatory. I also believe that using it inside the callbacks list is a better practice.

blumenstiel:
@romeokienzler @Joao-L-S-Almeida With this fix, the user cannot overwirte the default settings. So I don't think you should merge it with the current code. At least, check if the user defines a checkpoint and only add if it is missing. 
Also you need to test if the code is working if `ModelCheckpoint` is provided in the config, I would assume that is mailing because the argument is missing.

Joao-L-S-Almeida:
Right. We can do it.

Joao-L-S-Almeida:
The file `tests/resources/configs/manufactured-finetune_prithvi_eo_v2_300.yaml` was used for testing.

blumenstiel:
Thanks @Joao-L-S-Almeida, looks good. @romeokienzler Can you have another look and test it?

romeokienzler (approved):
lgtm

romeokienzler:
@blumenstiel seems your approval is needed as well

PR #457, opened 2025-02-25
Trying to solve issues with UNet and ASPPHeads. 

Joao-L-S-Almeida:
Alternative to #456.


terratorch/tasks/regression_tasks.py
```diff
@@ -292,7 +292,7 @@ def training_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) ->
         y = batch["mask"]
         other_keys = batch.keys() - {"image", "mask", "filename"}
         rest = {k: batch[k] for k in other_keys}
-
+        self.eval()
```
singhshraddha, line 295: I looked into this a bit and setting the whole model to eval also [affects the dropout layers](https://discuss.pytorch.org/t/will-eval-mode-disable-this-dropout/119073/2), so I don't think we can do this

A better way to just set batch norm layers to eval is mentioned in [this post](https://discuss.pytorch.org/t/should-i-use-model-eval-when-i-freeze-batchnorm-layers-to-finetune/39495/5). But, more importantly I don't think we should print a warning about using BatchNorm when batch size is 1 because it is also just a bad idea to do so

terratorch/models/decoders/aspp_head.py
```diff
@@ -255,6 +261,7 @@ def forward(self, inputs):
 
         return output
 
+@TERRATORCH_DECODER_REGISTRY.register
```
singhshraddha, line 264: This is actually for [line 276](https://github.com/IBM/terratorch/blob/b502907f4f6940bddfd0bcf96cca79ea188155f1/terratorch/models/decoders/aspp_head.py#L276) but I can only review changed lines

The way that `ASPPRegressionHead` is setup doesn't comply with the way [encoder_decoder_factory.py ](https://github.com/IBM/terratorch/blob/b502907f4f6940bddfd0bcf96cca79ea188155f1/terratorch/models/encoder_decoder_factory.py#L60) calls decoder build. Since, the build expects that the first arguments with be _channel_list_ and we have _dilations_ instead. So, the dilations are set incorrectly and I cannot set dilations using a config

To resolve this I'd suggest 
- not accepting `in_channels` as input parameter
- setting `channel_list` as the first input parameter
- Using `in_index` with `channel_list` to set the `in_channels`

```
class ASPPRegressionHead(ASPPHead):

    def __init__(
        self, 
        channel_list,
        dilations:list | tuple =(1, 6, 12, 18), 
        channels:int=None,
        out_channels:int=1,
        align_corners=False,
        head_dropout_ratio:float=0.3,
        input_transform: str = None,
        in_index: int = -1,
        **kwargs
):

   self.in_channels = channel_list[in_index]
``` 

**Note** this only works when input_transform is None. I currently only care about this use case wherein we select one item from the channel list and this corresponds to the item we pick from the dec_outs.  Though if you want to make it more generic mmseg has a `BaseDecodeHead._init_inputs` function that deals with different types of channel_list, input_transforms etc

terratorch/tasks/segmentation_tasks.py
```diff
@@ -243,7 +243,7 @@ def training_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) ->
         other_keys = batch.keys() - {"image", "mask", "filename"}
 
         rest = {k: batch[k] for k in other_keys}
-
+        self.eval()
```
singhshraddha, line 246: Same as the review in regression_tasks.py for setting model.eval()

singhshraddha (changes requested):

terratorch/models/backbones/unet.py
```diff
@@ -149,7 +152,7 @@ def __init__(self,
         self.strides = strides
         self.downsamples = downsamples
         self.norm_eval = norm_eval
-        self.base_channels = base_channels
+        self.out_channels = num_stages*[out_channels]
```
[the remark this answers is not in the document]
Joao-L-S-Almeida: Thanks. Just filling space to see things running.

terratorch/tasks/regression_tasks.py
Joao-L-S-Almeida: I believe that doesn't make much sense to use` batch_size=1` for models with batch normalization layers, even if we add a kind of workaround to make it run. 
Maybe the best solution is to include a global check, for any model in terratorch, which will raise and Exception for this cases, indicating the user to adopt `batch_size > 1` when BatchNorm is used. 
What do you think, @romeokienzler ?

terratorch/models/decoders/aspp_head.py
Joao-L-S-Almeida: Thanks. I'll try to adapt and include it in the ASPP module.

terratorch/tasks/segmentation_tasks.py
Joao-L-S-Almeida: I have removed all this calls. I belive the best approach it to block the usage of batch_size < 2. I'll do it in another PR.

singhshraddha (approved):
Looks fine for my use case. I'll run a model next week using this branch thanks

Joao-L-S-Almeida:
Hi, @singhshraddha Should we proceed and merge it ?

terratorch/cli_tools.py
```diff
@@ -102,7 +102,7 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16"):
     mask = np.where(mask == metadata["nodata"], 1, 0)
     mask = np.max(mask, axis=0)
     result = np.where(mask == 1, -1, prediction.detach().cpu())
-
+    print(result.shape)
```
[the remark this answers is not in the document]
Joao-L-S-Almeida: Done.

Joao-L-S-Almeida:
@singhshraddha All right. Let me know if you need any further modification. You also can use `mmseg` directly from teratorch as seen in `tests/resources/configs/manufactured-train_segmentation_unet.yaml`.

singhshraddha (approved):

PR #459, opened 2025-02-26
Basic support for visualizing models (as graphs)

Joao-L-S-Almeida:
It enables the visualization of models as large graphs (see `examples/examples/notebooks/Visualize_model.ipynb` in this branch)


Joao-L-S-Almeida:
As suggested in #435.

PR #461, opened 2025-02-27
[WiP] padding as input transform

Joao-L-S-Almeida:


blumenstiel:
@Joao-L-S-Almeida I think there is a `PadIfNeeded` in albumentations. Why are you not just using this one?

Joao-L-S-Almeida:
Simple. I never head about it before. @blumenstiel

Joao-L-S-Almeida:
@blumenstiel  In anyway. I also want to remove lines when necessary.

blumenstiel:
Haha good point:) Maybe test if it is working for you and replace the examples with it. I don't know in which version it was added, but it would be another reason to work on support for albumentation v2

Joao-L-S-Almeida:
@blumenstiel But the users can use PadIfNeeded if they want, or other equivalent operation. I think more than the transform itself the point here is to show them that this is a better option than checking stackability and modifying batch_norm.

PR #463, opened 2025-02-28
pinning albumentations==1.4.6

Joao-L-S-Almeida:
@blumenstiel has observed that D4 is missing in `albumentations==1.4.0`.


romeokienzler (approved):
lgtm

PR #483, opened 2025-03-07
allow users to run iterate using terratorch command

leotizzei:
iterate is a terratorch plugin, so the idea is to allow users to use terratorch command to call iterate: 
```shell
terratorch iterate --hpo --config ./configs/config.yaml
```


docs/quick_start.md
Joao-L-S-Almeida, line 1: It would be interesting to indicate that this is still in an experimental stage.

terratorch/__main__.py
Joao-L-S-Almeida, line 1: It's necessary to do a conditional import for `benchmark`. Maybe a try-except trying to import it and printing a logging message when that isn't installed.

Joao-L-S-Almeida: It would also be interesting to have an automatic test as those in `tests/test_finetune.py` specifically for `terratorch iterate`.

Joao-L-S-Almeida: We also have replaced print wherever possible with logging, as seen in `terratorch/models/backbones/prithvi_mae.py:29`.

Joao-L-S-Almeida: When iterate is not installed, we could instruct the user to do it via `pip install git+<repo_URL>`.

leotizzei: Hi Lucas,

1. conditional import -> done
2. automatic test: I'm not sure that we should include iterate test cases into terratorch tests, because I think it makes sense to test the code that is located in this terratorch's repo and not elsewhere.
3. use logging -> done
4. how to install iterate -> done, add a link on the docs to the installation section on terratorch-iterate

terratorch/__main__.py
```diff
@@ -2,11 +2,140 @@
...
 from terratorch.cli_tools import build_lightning_cli
+import sys
+from typing import List, Any
+try:
+    from benchmark.benchmark_types import Defaults, Task
+    from benchmark.backbone_benchmark import benchmark_backbone, rerun_best_from_backbone
+    from benchmark.utils import get_logger
+    TERRATORCH_ITERATE_INSTALLED = True
 
+except ImportError:
```
romeokienzler, line 18: @Joao-L-S-Almeida suggesting: if sys.argv[1] == "iterate" and TERRATORCH_ITERATE_INSTALLED = False print a message: please install iterate via pip install ....

leotizzei: done https://github.com/leotizzei/terratorch/blob/811509f0596d7368b9f81e147c8eb7b2d73f2ad9/terratorch/__main__.py#L23

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Feel free to merge,  @leotizzei .

PR #494, opened 2025-03-11
Updating the tests workflows to install granitewxc and prithviwxc from PyPI.

Joao-L-S-Almeida:
* They are available on PyPI, so it's better to use them instead of the GitHub versions. 
* Trying to remove the limitation over Python 3.12.


romeokienzler (approved):
lgtm

PR #500, opened 2025-03-12
Adjust in the checkpoint adjuster script

Joao-L-S-Almeida:


romeokienzler:
@paolo-fraccaro to test

PR #501, opened 2025-03-13
Update cli_tools.py, make subcommand optional

romeokienzler:


terratorch/cli_tools.py
Joao-L-S-Almeida, line 1: I understand it just would have effect if we run `terratorch` without any subcommand, but in this case, we have:
```
INFO:albumentations.check_version:A new version of Albumentations is available: 2.0.5 (you have 1.4.10). Upgrade using: pip install --upgrade albumentations
INFO:terratorch:Found custom_modules
usage: terratorch [-h] [-c CONFIG] [--print_config[=flags]] {fit,validate,test,predict,compute_statistics} ...
error: expected "subcommand" to be one of {fit,validate,test,predict,compute_statistics}, but it was not provided.

```

Joao-L-S-Almeida: I just saw that it could be necessary when running from the script interface, since this subcommand isn't passed in these cases.

PR #508, opened 2025-03-13
Options to directly informing kwargs for backbones and decoders

Joao-L-S-Almeida:
* Informing the arguments adding a prefix (`backbone_`, `decoder_`, `head_`) can work for most part of the cases, but for some specific uses, as arguments called `decoder_channels`, it will produce conflicts. 
* This solution adds the options of passing the arguments inside a dictionary, which will avoid this bad interpretation. 
* When using YAML files, the input will look like:
```
decoder: SomeInterestedDecoder
decoder_kwargs:
       decoder_channels: [100, 200, 300, 400]
```


romeokienzler (approved):
lgtm

PR #521, opened 2025-03-17
[WiP] Trying to merge with `main` before running tests

Joao-L-S-Almeida:
It tries to avoid the tests be executed again when the push is sent to `main` after a PR is approved.


Joao-L-S-Almeida:
Useless. no one cares.

Issue #460, opened 2025-02-27
Can't register a decoder which includes head


fmartiescofet:
**Describe the issue**
Right now the value of `includes_head` is defined at registry level. I would suggest making it at decoder level as each decoder could be different, for example we may want to create a new decoder in the `TERRATORCH_DECODER_REGISTRY` which already includes the head and this is not possible with the current setup. This is because [this line](https://github.com/IBM/terratorch/blob/92bd85a16cb7301d90f6b674d32aadfba83b079e/terratorch/models/encoder_decoder_factory.py#L50) is checked at a registry level.

romeokienzler:
@fmartiescofet saying it got solved with 1.0 #527

PR #527, opened 2025-03-19
Add linear decoder and check `includes_head` at decoder level first

fmartiescofet:
This PR introduces a `LinearDecoder`, which can be particularly useful for experiments involving minor modifications to the backbone. By using a lightweight decoder, we can better isolate the impact of these changes without the influence of a larger, more complex decoder.
It also solves #460 by checking the `includes_head` at the decoder level and then at the registry level.


terratorch/models/decoders/linear_decoder.py
Joao-L-S-Almeida, line 1: Annotations are missing.

Joao-L-S-Almeida:
Thanks @fmartiescofet. It's approved and you can merge when you prefer.

fmartiescofet:
Thanks @Joao-L-S-Almeida, I don't have merge rights, feel free to merge!

PR #528, opened 2025-03-19
Implement ViT-Adapter

fmartiescofet:
This PR implements the [ViT-Adapter](https://github.com/czczup/ViT-Adapter) method for Prithvi backbones. It works only with non-multitemporal dataset. The issue is we can't run tests on github due to the fact that it needs CUDA to use deformable attention.

@blumenstiel @romeokienzler


terratorch/models/backbones/prithvi_vit_adapter.py
Joao-L-S-Almeida, line 1: Docstrings.

terratorch/models/backbones/detr_ops/modules/ms_deform_attn.py
Joao-L-S-Almeida, line 1: Docstrings in Google style.

terratorch/models/backbones/detr_ops/functions/ms_deform_attn_func.py
Joao-L-S-Almeida, line 1: As you said, it depends on CUDA devices, which could difficult the automatic tests on a GH Actions, but we can run it for integrations tests executed in a dedicated device, so we don't need to care about it.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -1,15 +1,30 @@
...
+
 from terratorch.datasets import HLSBands
-from terratorch.models.backbones.select_patch_embed_weights import select_patch_embed_weights
 from terratorch.datasets.utils import generate_bands_intervals
-from terratorch.models.backbones.prithvi_mae import PrithviViT, PrithviMAE
+from terratorch.models.backbones.prithvi_mae import PrithviMAE, PrithviViT
+from terratorch.models.backbones.select_patch_embed_weights import select_patch_embed_weights
 from terratorch.registry import TERRATORCH_BACKBONE_REGISTRY
-from huggingface_hub import hf_hub_download
+
```
Joao-L-S-Almeida, line 15: Why is this conditional  necessary in this case ?  If we are running a version of terratorch containing `PrithviViTAdapter` it should be always available, right ? 
In any case, it would be helpful to have a warning message about the lack of this model.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -221,15 +339,18 @@ def _create_prithvi(
...
+                f"No pre-trained model found for variant {variant} (pretrained models: {pretrained_weights.keys()})"
+            )
 
             try:
                 # Download config.json to count model downloads
                 _ = hf_hub_download(repo_id=pretrained_weights[variant]["hf_hub_id"], filename="config.json")
                 # Load model from Hugging Face
-                pretrained_path = hf_hub_download(repo_id=pretrained_weights[variant]["hf_hub_id"],
-                                                  filename=pretrained_weights[variant]["hf_hub_filename"])
+                pretrained_path = hf_hub_download(
```
Joao-L-S-Almeida, line 350: Just reformatted.

Joao-L-S-Almeida, line 350: Just reformatted.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -257,137 +384,162 @@ def forward_filter_indices(*args, **kwargs):
...
+) -> PrithviViT | PrithviViTAdapter:
+    warnings.warn(
+        f"The model prithvi_vit_tiny was renamed to prithvi_eo_tiny. "
+        f"prithvi_vit_tiny will be removed in a future version.",
+        FutureWarning,
+    )
 
-    warnings.warn(f"The model prithvi_vit_tiny was renamed to prithvi_eo_tiny. "
-                  f"prithvi_vit_tiny will be removed in a future version.", FutureWarning)
+    return prithvi_eo_tiny(pretrained=pretrained, vit_adapter=vit_adapter, **dict({"model_bands": bands}), **kwargs)
```
Joao-L-S-Almeida, line 473: vit_adapter is False by default.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -257,137 +384,162 @@ def forward_filter_indices(*args, **kwargs):
...
+        FutureWarning,
+    )
 
-    return prithvi_eo_v1_100(pretrained=pretrained, **dict({"model_bands": bands}, **kwargs))
+    return prithvi_eo_v1_100(pretrained=pretrained, vit_adapter=vit_adapter, **dict({"model_bands": bands}), **kwargs)
 
 
 # TODO: Remove timm_ errors before version v1.0.
-@ TERRATORCH_BACKBONE_REGISTRY.register
+@TERRATORCH_BACKBONE_REGISTRY.register
```
Joao-L-S-Almeida, line 493: From here to the bottom, just reformatted. Right.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -41,53 +56,103 @@ def _cfg(**kwargs):
...
+        num_heads=16,
+        patch_size=[1, 14, 14],
+        coords_encoding=["time", "location"],
+        coords_scale_learn=True,
+    ),
 }
 
 # Timm pretrained configs
 pretrained_weights = {
-        "prithvi_eo_v1_100": {
```
Joao-L-S-Almeida, line 70: Just changed indentation. Right.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -41,53 +56,103 @@ def _cfg(**kwargs):
...
         "coords_scale_learn": False,
         "bands": PRETRAINED_BANDS,
         "mask_ratio": 0.75,
         "norm_pix_loss": False,
-        **kwargs
+        **kwargs,
     }
 
 
 prithvi_cfgs = {
```
Joao-L-S-Almeida, line 70: Were these entries not being used  from `prithvi_cfgs` ? In this way you removed them ?
In fact, in `terratorch:main` this dictionary is just used in `_create_prithvi`, where you partially replaced it with this new `prithvi_adapter_cfgs`. 
Why just these entries are managed for ViT-adapter and why `prithvi_eo_v2_600` and `prithvi_eo_v2_600_tl` aren't moved to this new dict too ?

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -41,53 +56,103 @@ def _cfg(**kwargs):
...
     "prithvi_eo_v1_100": _cfg(num_frames=3, mean=PRITHVI_V1_MEAN, std=PRITHVI_V1_STD),
     "prithvi_eo_v2_100": _cfg(),
     "prithvi_eo_v2_300": _cfg(embed_dim=1024, depth=24, num_heads=16),
-    "prithvi_eo_v2_300_tl": _cfg(embed_dim=1024, depth=24, num_heads=16,
-                                 coords_encoding=["time", "location"], coords_scale_learn=True),
+    "prithvi_eo_v2_300_tl": _cfg(
+        embed_dim=1024, depth=24, num_heads=16, coords_encoding=["time", "location"], coords_scale_learn=True
+    ),
     "prithvi_eo_v2_600": _cfg(embed_dim=1280, depth=32, num_heads=16, patch_size=[1, 14, 14]),
-    "prithvi_eo_v2_600_tl": _cfg(embed_dim=1280, depth=32, num_heads=16, patch_size=[1, 14, 14],
```
Joao-L-S-Almeida, line 64: Just reformatted. Right.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -41,53 +56,103 @@ def _cfg(**kwargs):
...
         "norm_pix_loss": False,
-        **kwargs
+        **kwargs,
     }
 
 
 prithvi_cfgs = {
-    "prithvi_eo_tiny": _cfg(num_frames=1, embed_dim=256, depth=4, num_heads=4,
-                            decoder_embed_dim=128, decoder_depth=4, decoder_num_heads=4),
+    "prithvi_eo_tiny": _cfg(
```
Joao-L-S-Almeida, line 71: Just reformatted. Right.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -41,53 +56,103 @@ def _cfg(**kwargs):
...
+        num_frames=1, embed_dim=256, depth=4, num_heads=4, decoder_embed_dim=128, decoder_depth=4, decoder_num_heads=4
+    ),
     "prithvi_eo_v1_100": _cfg(num_frames=3, mean=PRITHVI_V1_MEAN, std=PRITHVI_V1_STD),
     "prithvi_eo_v2_100": _cfg(),
     "prithvi_eo_v2_300": _cfg(embed_dim=1024, depth=24, num_heads=16),
-    "prithvi_eo_v2_300_tl": _cfg(embed_dim=1024, depth=24, num_heads=16,
```
Joao-L-S-Almeida, line 61: Reformatted. Right.

terratorch/models/backbones/prithvi_vit.py
```diff
@@ -41,53 +56,103 @@ def _cfg(**kwargs):
...
-                                 coords_encoding=["time", "location"], coords_scale_learn=True),
+    "prithvi_eo_v2_300_tl": _cfg(
+        embed_dim=1024, depth=24, num_heads=16, coords_encoding=["time", "location"], coords_scale_learn=True
+    ),
     "prithvi_eo_v2_600": _cfg(embed_dim=1280, depth=32, num_heads=16, patch_size=[1, 14, 14]),
```
Joao-L-S-Almeida, line 80: Removed ?

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: Yes

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: I didn't remove them, it just got reformatted when I saved, sorry about that. These configs are still used in the same way they were before, now I just added the `prithvi_adapter_cfgs` dictionary which contains the extra arguments needed to build the `PrithviViTAdapter` for some models. Also the 600M models are not implemented because in the original paper they don't use them and therefore I didn't create configs for them. They only implement the adapter for ViT-B and ViT-L.

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: Yes

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: Yes

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: No, see comment before.

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: Yes

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: Yes

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: Yes, it's always `False` by default. The user needs to explicitly pass it to use it.

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: Yes.

terratorch/models/backbones/prithvi_vit.py
fmartiescofet: It's a little bit weird, the issue comes from the fact that if the use does not have deformable attention installed then there is an `ImportError` when trying to define it (because it's needed for the `PrithviViTAdapter`). Therefore I added this check, now I refactored and always define it in `prithvi_vit_adapter.py` and use a dummy class which raises an error if trying to instantiate it. I don't think we should add a warning because most of the users won't have deformable attention installed and therefore it will practically always be shown. I believe it's best to raise an error if a user tries to use it and it's not installed.

Joao-L-S-Almeida (approved):

PR #534, opened 2025-03-20
Handling many possibilities when trying to perform the test step with support for tiled inference

Joao-L-S-Almeida:


terratorch/tasks/base_task.py
```diff
@@ -50,6 +54,64 @@ def configure_models(self) -> None:
...
+            # When the user don't set the variable `tiled_inference_on_testing`
+            # as `True` in the config, we will try to use full inference.
+            try:
+                model_output: ModelOutput = self(x, **rest)
+            except (torch.OutOfMemoryError, MemoryError)  as e:
+                raise Exception("Inference on testing failed due to memory issues. Try to pass `tiled_inference_on_testing` as `True`, to use tiled inference for it.")
+
+        else:
+            try:
+                # Even when `tiled_inference_on_testing` is `True`, we at least
```
romeokienzler, line 80: I would not do that. This is a bit of black magic. The framework is doing things which the user not explicitly requested to do. So users could get confused if they think they run tiled inference on one system but in the background it's not and are happy with the model performance, then on a different system it will run into the OOM error and deliver inferior results 

@paolo-fraccaro @Joao-L-S-Almeida agree?

romeokienzler (changes requested):
@Joao-L-S-Almeida there is one commend we should discuss, rest lgtm, thanks a lot!

romeokienzler (approved):
Lgtm thanks so much!

PR #538, opened 2025-03-25
improve_wxc_import_error_handling

romeokienzler:
Log import error on debug level during init (for all users), raise error in case wxc packages are not installed but user is trying to use


blumenstiel (approved):
Thanks!

PR #542, opened 2025-03-26
Correct number of bands for  single band (as for segmentation) output file. 

Joao-L-S-Almeida:
The `count` parameter was being modified inside `write_tiff`, making the output files with an artificial number of bands.


romeokienzler (approved):
lgtm

PR #550, opened 2025-03-27
support / accept contributions according to MIT license

romeokienzler:


Joao-L-S-Almeida:
@romeokienzler  I understood that you indicate Apache as out core license and appended a part indicating that some modules are licensed by MIT. Are you going to add the file  `MIT_FILES.txt` ? I don't know to map the regions of our code which are licensed under MIT.

Joao-L-S-Almeida (approved):

romeokienzler:
@Joao-L-S-Almeida yes I'll create a script to search for similar code regions against different candidate repositories

PR #560, opened 2025-03-31
Testing for python 3.13

Joao-L-S-Almeida:


Joao-L-S-Almeida:
The tests are taking too much time. We need to simplify them and improve their efficiency before testing for more Python versions.

.github/workflows/pylint.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10","3.11","3.12"]
+        python-version: ["3.10","3.11","3.12","3.13"]
```
romeokienzler, line 14: shall we remove 3.10?

romeokienzler (approved):

Joao-L-S-Almeida:
The job is suddenly interrupted without any clear reason. Running the tests locally I have seen that they work, but using sometimes more than 16 GB of memory. maybe that is the reason for the interruption.

Joao-L-S-Almeida:
It seems that doesn't matter what we did, the operation is cancelled in some moment.

Joao-L-S-Almeida:
For some reason, this is now working, @romeokienzler .

PR #567, opened 2025-04-01
Guaranteeing that the checkpointing operations are understood by Lightning

Joao-L-S-Almeida:
The dictionaries were replaced with `jsonargparse._namespace.Namespaces` to avoid any confusion during the instantiation. 
@romeokienzler @fmartiescofet, check if you want others arguments are added to default [here](https://github.com/IBM/terratorch/blob/a22a95c5afb29d52ad37d221f9f456a8a2f373fa/terratorch/cli_tools.py#L124).

### DIFF
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -28,6 +28,7 @@ import torchgeo.datamodules
 import yaml
 from albumentations.pytorch import ToTensorV2  # noqa: F401
 from jsonargparse import set_dumper
+from jsonargparse._namespace import Namespace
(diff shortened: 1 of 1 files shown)


fmartiescofet:
I tested it and it looks like it's working correctly now and the callbacks are used. Thanks!

Joao-L-S-Almeida:
Thanks, @fmartiescofet.

romeokienzler (approved):
lgtm

@Joao-L-S-Almeida I think the defaults you've chosen are a great choice

PR #569, opened 2025-04-02
[WIP] add json schema for configs

romeokienzler:


romeokienzler:
@Joao-L-S-Almeida all config files than wxc ones are passing, so I think we can merge, the wxc ones have a problem, so will fix there

Joao-L-S-Almeida (approved):

PR #577, opened 2025-04-03
Update DOFA Backbone for Arbitrary Image Sizes

xiong-zhitong:
### What’s Changed
- Updated the DOFA backbone to support arbitrary input image sizes by interpolating position embeddings.
This enhancement is critical for dense prediction tasks such as semantic segmentation and object detection, where higher-resolution inputs significantly improve performance.

- Corrected the wavelength setting for Sentinel-1 SAR images in the updated version of DOFA weights.

- Added an explicit link to the DOFA pre-trained weights to ensure better control over which version is being used.


paolofraccaro:
@xiong-zhitong this is great thanks for submitting this! All looks good! My only comment is: do you want to maybe expose the interpolation method for the pos embeddings to the user? From the code seems to be fixed on `bilinear` but more options seem to be available.

paolofraccaro (approved):
Thanks @xiong-zhitong ! This looks good to me!

paolofraccaro (changes requested):
I just realised something: can we remove the object detection config from the PR until we merge object detection in main? The DOFA changes are all good and we should include!

xiong-zhitong:
Should we also merge this update to the main branch?

PR #578, opened 2025-04-03
Recording metrics to a CSV file

Joao-L-S-Almeida:


terratorch/tasks/base_task.py
```diff
@@ -156,3 +160,16 @@ def _do_plot_samples(self, batch_index):
...
+        if self.path_to_record_metrics:
+            # Recording the metrics
+            metrics_record_ = self.test_metrics[dataloader_idx](y_hat_hard, y)
+            metrics_record_dict = {k: float(np.array(v.detach().cpu())) for k,v in metrics_record_.items()} 
+            metrics_record_list = [{"Metric": k, "Value": v} for k, v in metrics_record_dict.items()]
+            metrics_record = pd.DataFrame(data=metrics_record_list)
+
+            filename = os.path.join(self.path_to_record_metrics, "metrics.csv")
+            metrics_record.to_csv(filename)
+
```
romeokienzler, line 175: @Joao-L-S-Almeida just to make the code safer I'd throw an exception in case path_to_record_metrics is empty (failfast)

romeokienzler: sorry @Joao-L-S-Almeida just realized that you are pushing the decision logic into the function, which is a different style I'm used to but I like it as the decision logic does not need to be duplicated at caller side and also reduces control flow, thanks!!!

tests/resources/configs/manufactured-finetune_prithvi_eo_v2_600_segmentation_tiled.yaml
```diff
@@ -101,6 +101,7 @@ model:
       head_dropout: 0.5708022831486758
     output_most_probable: false
     tiled_inference_on_testing: true
+    path_to_record_metrics: ./
```
romeokienzler, line 104: @Joao-L-S-Almeida we need - if we add functionality to the config file - always update 
- documentation
- json schema

Joao-L-S-Almeida: The documentation will be updated for `stable` as soon as the PR is approved. The documentation is always generated from the  latest `main` branch. 
I think we need to include more conditionals to the json scheme. These arguments are specific for segmentation.

romeokienzler (approved):

PR #599, opened 2025-04-09
Organise examples and update links

paolofraccaro:


examples/notebooks/WxCTutorialDownscaling.ipynb
```diff
@@ -1,3642 +0,0 @@
-{
```
romeokienzler, line 1: why have those been deleted without replacement @paolofraccaro ?

paolofraccaro: my bad sorry!

PR #601, opened 2025-04-09
Update get_common_prefix to handle cases where len(prefix_list) < 1

fgiauna:
While working with PrithviMAE by setting `"backbone_encoder_only" = False` I got an error with the function `get_common_prefix` from models/backbone/select_patch_embed_weights.py.
There was no common prefix in my state_dict so prefix_list was an empty set and the else condition led to :
> KeyError: 'pop from an empty set'

Moreover I think that the intended return statement is `prefix` and not `prefix + "."`.

I also removed the argument `infer_datetime_format=True` in datasets/multi_temporal_crop_classification.py as it spams warnings.
infer_datetime_format is deprecated since pandas version 2.0.0 as stated in the docs : [pandas.to_datetime](https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html)

### DIFF
--- a/terratorch/datasets/multi_temporal_crop_classification.py
+++ b/terratorch/datasets/multi_temporal_crop_classification.py
@@ -157,7 +157,7 @@ class MultiTemporalCropClassification(NonGeoDataset):
         temporal_coords = []
         for col in self.date_columns:
             date_str = row[col]
-            date = pd.to_datetime(date_str, infer_datetime_format=True)
+            date = pd.to_datetime(date_str)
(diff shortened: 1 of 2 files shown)


Joao-L-S-Almeida:
Hi, @fgiauna Thank you for this contribution. 
I'd like to reproduce your issue. Could you provide us a snippet of code with the test you tried to do ? 
We can also extend our automatic tests to include it.

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
@fgiauna Thanks. I agree with your changes.

PR #606, opened 2025-04-11
A better way to remove the prefix `model.` from some old checkpoints. 

Joao-L-S-Almeida:
This line was replacing `model` in others regions of the key string, then producing errors when loading specific checkpoints. The loop combined with a conditional will avoid it. 

For example, in the previous version a key named `model.smp_model.decoder.blocks.1.conv1.1.num_batches_tracked` should be renamed to `smp_model.decoder.blocks.1.conv1.1.num_batches_tracked`, but it is erroneously renamed to `smp_decoder.blocks.1.conv1.1.num_batches_tracked`.


Joao-L-S-Almeida:
Thanks.

PR #607, opened 2025-04-14
Predownload model configs

romeokienzler:


terratorch/__main__.py
Joao-L-S-Almeida, line 1: Is this command allowing the user run `terratorch init <args>` and it will download and prepare the configs ?

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Approved. Lgtm.

PR #617, opened 2025-04-17
Imagefolder/dummy tiff file

Joao-L-S-Almeida:
There was an issues when the user generated a floating-point image. The images were recorded and loaded as tensors of zeroes. 
* The new dummy image generator will handle floating-point cases.
* The new dataset `__getitem__` method will manage when load floating-point images without truncating them to zero.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ dependencies = [
   "h5py",
   "mlflow",
   "lightning",
-  "segmentation-models-pytorch",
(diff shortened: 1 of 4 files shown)


PedroConrado (approved):

Joao-L-S-Almeida:
@romeokienzler and @PedroConrado It will need be re-approved after the fix for the tests.

PedroConrado (approved):

PR #635, opened 2025-04-23
Periodical synchronization for the documentation.

Joao-L-S-Almeida:
The documentation that is deployed in the dedicated branch `improve/docs` must be periodically synchronized with the `main`.


terratorch/models/generic_model_factory.py
```diff
@@ -25,25 +24,17 @@ class GenericModelFactory(ModelFactory):
 
     def build_model(
         self,
-        task: str = "segmentation",
```
romeokienzler, line 28: why are those default parameters removed?

Joao-L-S-Almeida: Yes, they aren't used, so I decided to remove before building the latest docs.

PR #636, opened 2025-04-23
It's better to have default values for `img_grep` and `label_grep`

Joao-L-S-Almeida:
With a missing default value there is an error when the user don't provide values in the config.


.github/workflows/test.yaml
```diff
@@ -39,6 +39,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -e .[wxc,test]
+      - name: Clean pip cache
```
romeokienzler, line 42: again, but should not be a problem...

romeokienzler (approved):

PR #645, opened 2025-04-25
Fix predict and tiled inference

blumenstiel:
- Fix error with generic data modules when using `predict`
- Make batch size configurable for tiled inference
- Moved pred in tiled inference to cpu to avoild OOM errors


blumenstiel:
I also added a `verbose` setting to see the progress for very large tiles.

Updated settings:
```
    tiled_inference_parameters:
        ...
        batch_size: 16  # default
        verbose: false # default
```

blumenstiel:
@romeokienzler @paolo-fraccaro @Joao-L-S-Almeida We have the issue that this still leads to OOM errors if the input tile is very large (e.g. 100k x100k). My assumption is that Lightling loads the input to the GPU device during data loading which already leads to the OOM error. So copying it back to CPU in tiled inference does not work. Any idea how we can avoild this without changing to much code?

cc @jhnnsjkbk1

blumenstiel:
@Joao-L-S-Almeida Thanks for updating this pr! Do you think it is ready for review by @romeokienzler?

romeokienzler:
@blumenstiel on it - just wanted to run the integration tests but have a glitch there...

romeokienzler (approved):

Issue #655, opened 2025-04-30
custom sequential schedulers


naomi-simumba:
**Is your feature request related to a problem? Please describe.**
current implementation of `optimizer_factory`  only allows for optimisers from the `torch.optim.lr_scheduler` class. 
https://github.com/IBM/terratorch/blob/25fc5e6969d680edd8e5e779f98d6a2744b69c5d/terratorch/tasks/optimizer_factory.py#L9

More complex scheduling may be achieved with the SequentialLR scheduler, which uses multiple schedulers. However, this is cannot be used with the current set-up.

**Describe the solution you'd like**
Add functionality to unpack Sequential schedulers. e.g. <link> 

**Describe alternatives you've considered (optional)**
Users create custom tasks. This would mean copying the entire existing task and changing only the `configure_optimizers` fn.

**Additional context (optional)**
Several newer schedulers are being used in the community and are not available from `torch`. However, these may be available from other libraries. e.g. `timm`.

PR #656, opened 2025-04-30
unpack sequential scheduler

naomi-simumba:
Addresses the following issues:
 https://github.com/IBM/terratorch/issues/655
https://github.com/IBM/terratorch/issues/696


terratorch/tasks/optimizer_factory.py
Joao-L-S-Almeida, line 1: You added a special condition when the user selects `SequentialLR` in which you update the dictionary `scheduler_hparams_no_interval` with extra parameters (whose the existence is asserted some line above). 
All right.

terratorch/models/smp_model_factory.py
Joao-L-S-Almeida, line 1: I understand you moved the previous block of code to `else` and added a special case for backbone names starting with `tu-`.  What is special with this class of models ?

Joao-L-S-Almeida:
@naomi-simumba, I agree with the changes, but I think it would be important to add a few automatic tests under `tests/`.

terratorch/models/smp_model_factory.py
naomi-simumba: These are technically backbone from the `timm` library which are not natively available in the `SMP` library, meaning they are not in `smp_encoders`. The `tu-` prefix tells SMP to load the model directly from `timm`

terratorch/tasks/optimizer_factory.py
naomi-simumba: Yes, the extra parameters added to `scheduler_hparams_no_interval` are schedulers that must be executed in sequence

naomi-simumba:
> @naomi-simumba, I agree with the changes, but I think it would be important to add a few automatic tests under `tests/`.

I have added some tests to the latest commit

PR #659, opened 2025-05-01
[WiP] Onboarding Galileo

Joao-L-S-Almeida:


Joao-L-S-Almeida:
@paolo-fraccaro I'm asking about the future of this PR. Galileo is now accessible and there is a decoder adapted from the original decoder used for pre-training. However, it's unclear what is the best way of using it, or if using a simple reshaping neck is the best approach to adapt its outputs to be used with our examples. 
I would be helpful to get some feedback from the developers and interested groups.

Joao-L-S-Almeida:
@paolo-fraccaro I'll leave it for `1.2` until further notice.

Joao-L-S-Almeida:
Check test `tests/test_backbones.py::test_galileo_encoders` to see the combination with a `FCNDecoder`.

Joao-L-S-Almeida:
It seems the GalileoWrapper expects all the bands be filled for Sentinel-2. Don't we have the option of providing inputs with missing bands ?

pyproject.toml
```diff
@@ -106,6 +106,10 @@ peft = [
   "peft>=0.15.0"
 ]
```
romeokienzler, line 108: @Joao-L-S-Almeida wouldn't this prevent us from creating a release?

romeokienzler (approved):

pyproject.toml
Joao-L-S-Almeida: I don't think so. It is already like this in the `main`.

PR #666, opened 2025-05-05
Inference integration tests complete &updated eo v1 model paths

WanjiruCate:
Added tests for Inference part.

### DIFF
--- a/integrationtests/test_base_set.py
+++ b/integrationtests/test_base_set.py
@@ -60,7 +60,7 @@ def floods_image(tmp_path_factory):
     return str(local_path)
 
 
-def test_eo_v2_100_burns_fit():
+def test_eo_v1_100_burns_fit():
     # Call the CLI program
     result = subprocess.run(
         ['terratorch', 'fit', '-c', 'test_encoderdecoder_eo_v1_100_model_factory.yaml'], capture_output=True, text=True
@@ -76,7 +76,7 @@ def test_eo_v2_100_burns_fit():
     ), f"Test failed with return code {result.returncode}\nSTDOUT: {result.stdout}\nSTDERR: {result.stderr}"
 
 
-def test_eo_v2_100_buildings_predict(buildings_image, base_models_path):
+def test_eo_v1_100_buildings_predict(buildings_image, base_models_path):
     config_path = f"{base_models_path}buildings_eo_v1_100/config_eo_100.yaml"
     checkpoint_path = f"{base_models_path}buildings_eo_v1_100/checkpoint_eo_100.ckpt"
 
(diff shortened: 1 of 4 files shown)


integrationtests/test_base_set.py
```diff
@@ -50,14 +55,15 @@ def floods_image(tmp_path_factory):
...
+    download_and_open_tiff(url=url, dest_path=local_path)
 
     return str(local_path)
 
+
 def test_eo_v2_100_burns_fit():
     # Call the CLI program
     result = subprocess.run(
-        ['terratorch', 'fit', '-c', 'test_encoderdecoder_eo_v2_100_model_factory.yaml'], capture_output=True, text=True
+        ['terratorch', 'fit', '-c', 'test_encoderdecoder_eo_v1_100_model_factory.yaml'], capture_output=True, text=True
```
romeokienzler, line 66: why is the test function's name v2 and the yaml file name v1?

integrationtests/test_base_set.py
```diff
@@ -22,13 +22,18 @@ def run_inference(config, checkpoint, image):
     return predictions
 
 
+@pytest.fixture(scope="session")
```
romeokienzler, line 25: is there a reason for using a function for a constant?

integrationtests/test_base_set.py
```diff
@@ -70,6 +76,15 @@ def test_eo_v2_100_burns_fit():
     ), f"Test failed with return code {result.returncode}\nSTDOUT: {result.stdout}\nSTDERR: {result.stderr}"
 
 
+def test_eo_v2_100_buildings_predict(buildings_image, base_models_path):
```
romeokienzler, line 79: also name/file version mismatch?

integrationtests/test_encoderdecoder_eo_v2_300_model_factory.yaml
```diff
@@ -27,8 +27,6 @@ trainer:
         monitor: val/loss
         filename: best-state_dict-{epoch:02d}
         save_weights_only: True
-      
-  max_epochs: 2
```
romeokienzler, line 31: why has this been removed? it will just run longer, correct? or are you checking on the correct result after the run?

integrationtests/test_base_set.py
WanjiruCate: I wanted to reuse the base path across the different test functions.

integrationtests/test_encoderdecoder_eo_v2_300_model_factory.yaml
WanjiruCate: It was a duplicate of line 10.

integrationtests/test_base_set.py
WanjiruCate: Fiixed

integrationtests/test_base_set.py
romeokienzler: @WanjiruCate just wondering why you are not using a constant, but as it works fine we can also leave it like this

romeokienzler (approved):

PR #675, opened 2025-05-07
Adapting multimodal dataset to support empty label roots.

Joao-L-S-Almeida:
It fixes issues observed for prediction using the multimodal datasets.

### DIFF
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -212,6 +212,12 @@ class CustomWriter(BasePredictionWriter):
 
             pred_batch_, filename_batch = prediction
 
+            # In case of multimodal datasets, we have a dictionary
+            # with one file per modality
+            if type(filename_batch)==dict:
+                keys = list(filename_batch.keys())
+                filename_batch = filename_batch[keys[0]]
+
             # If we have just a single kind of output
             if isinstance(pred_batch_, tuple):
 
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -135,7 +135,7 @@ def add_default_checkpointing_config(config):
     return config 
 
 def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16",
-                    suffix="pred"):
+                    suffix:str="pred", output_file_name:str | None=None):
     mask, metadata = open_tiff(input_file_name)
     mask = np.where(mask == metadata["nodata"], 1, 0)
     mask = np.max(mask, axis=0)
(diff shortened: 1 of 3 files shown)


biancazadrozny (dismissed):

terratorch/cli_tools.py
```diff
@@ -147,7 +147,12 @@ def save_prediction(prediction, input_file_name, out_dir, dtype:str="int16",
     metadata["compress"] = "lzw"
     metadata["nodata"] = -1
     file_name = os.path.basename(input_file_name)
-    file_name_no_ext = os.path.splitext(file_name)[0]
+
+    if not output_file_name:
+        file_name_no_ext = os.path.splitext(file_name)[0]
+    else:
+        file_name_no_ext = output_file_name
```
romeokienzler, line 154: @Joao-L-S-Almeida so u expect the output_file_name always to be passed without extension?

terratorch/cli_tools.py
```diff
@@ -212,6 +217,12 @@ def write_on_batch_end(self, trainer, pl_module, prediction, batch_indices, batc
 
             pred_batch_, filename_batch = prediction
 
+            # In case of multimodal datasets, we have a dictionary
+            # with one file per modality
+            if type(filename_batch)==dict:
+                keys = list(filename_batch.keys())
+                filename_batch = filename_batch[keys[0]]
```
romeokienzler, line 224: @Joao-L-S-Almeida so we always take the file name of the first modality as filename_batch, is this how it is intended?

terratorch/datamodules/generic_multimodal_data_module.py
```diff
@@ -353,6 +360,16 @@ def __init__(
         self.val_label_data_root = val_label_data_root
         self.test_label_data_root = test_label_data_root
         self.predict_root = predict_data_root
+
```
romeokienzler, line 363: @Joao-L-S-Almeida can we agree to move away from assetions and raise a rte instead, please see https://realpython.com/python-raise-exception

romeokienzler (approved):

terratorch/cli_tools.py
Joao-L-S-Almeida: Yes, just a name. That isn't still used.

terratorch/cli_tools.py
Joao-L-S-Almeida: Yes, this is an workaround, the way we save tiff images is biased to single inputs. I'll modify it.

terratorch/datamodules/generic_multimodal_data_module.py
Joao-L-S-Almeida: Yes.

terratorch/cli_tools.py
Joao-L-S-Almeida: See commit 0b13e780bccb0ed026b93e51fc62d079ecac7183

terratorch/cli_tools.py
Joao-L-S-Almeida: See commit 0b13e780bccb0ed026b93e51fc62d079ecac7183

romeokienzler (approved):

PR #681, opened 2025-05-12
GPU support in Dockerfile, trigger nightly build

romeokienzler:


Joao-L-S-Almeida (approved):

.github/workflows/container_nightly.yaml
Joao-L-S-Almeida, line 1: Is this image being uploaded each time `main` is updated ?

Joao-L-S-Almeida (approved):

.github/workflows/container_nightly.yaml
romeokienzler: @Joao-L-S-Almeida yes - I was also hesitant, but main isn't updated that often and so far it seems to be fine - we can deactivate this and remove stream and have nightly only if it causes problems ... stream does only make sense if we continuously build, one idea is to trigger an integration test on a gpu cluster when ever stream is updated...that's why

PR #690, opened 2025-05-15
Periodical documentation synchronization with  `main`

Joao-L-S-Almeida:


blumenstiel:
Makes sense IMHO, especially before the update-docs pr. But I cannot tell if it looks alright, @romeokienzler?

PR #697, opened 2025-05-22
Update default metrics

blumenstiel:
- Updated to commonly used metrics (e.g. added precision and recall to the classification tasks).
- Metrics are calculated as macro by default, if not clearly mentioned in the name.
- Renamed Jaccard to IoU.


romeokienzler (approved):
tnx @blumenstiel 

- [ ] merged

PR #700, opened 2025-05-26
add era5 test

romeokienzler:


tests/datamodules/test_era5.py
```diff
@@ -0,0 +1,109 @@
...
+
+    assert batch["x"].shape == (2, 1, 488, 64, 128)
+    assert batch["y"].shape == (2, 488, 64, 128)
+    assert batch["lead_time"].shape == (2, 1)
+    assert batch["static"].shape == (2, 3, 64, 128)
+    gc.collect()
+
+    try:
+        os.remove('era5dummy.nc')
+        print("Dummy ERA5 dataset removed after test")
```
blumenstiel, line 104: See above

blumenstiel (changes requested):
Looks good, except for the print calls. Which model is using this module? Might be good to test it as well

blumenstiel (approved):

PR #713, opened 2025-06-06
add missing import: from torchmetrics.wrappers import ClasswiseWrapper

leotizzei:
it seems that the ClasswiseWrapper class has not been properly imported in `terratorch/tasks/multilabel_classification_tasks.py` so it raises this error message:
```shell
NameError: name 'ClasswiseWrapper' is not defined
```

### DIFF
--- a/terratorch/tasks/multilabel_classification_tasks.py
+++ b/terratorch/tasks/multilabel_classification_tasks.py
@@ -13,7 +13,7 @@ from torchmetrics.classification import (
     MultilabelRecall,
     MultilabelAUROC,
 )
-
+from torchmetrics.wrappers import ClasswiseWrapper
 from terratorch.models.model import ModelOutput
 from terratorch.tasks import ClassificationTask
 


bretcarpenter:
It looks like the `NameError: name 'ClasswiseWrapper' is not defined` is happening because the `ClasswiseWrapper` class isn’t being imported in **`terratorch/tasks/multilabel_classification_tasks.py`**. Python simply can’t find it, so the fix is to explicitly import it at the top of the file alongside the other TorchMetrics imports.

```diff
--- a/terratorch/tasks/multilabel_classification_tasks.py
+++ b/terratorch/tasks/multilabel_classification_tasks.py
@@ -1,6 +1,7 @@
 from torchmetrics import MetricCollection
+from torchmetrics.wrappers import ClasswiseWrapper
 from torchmetrics.classification import MulticlassAccuracy
 from torch import nn
 import pytorch_lightning as pl
```

With that import in place, you can initialize your metrics inside the task class without triggering a NameError. For example:

```python
from torchmetrics import MetricCollection
from torchmetrics.classification import MulticlassAccuracy
from torchmetrics.wrappers import ClasswiseWrapper

class MultiLabelClassificationTask(pl.LightningModule):
    def __init__(self, num_classes, class_names, ...):
        super().__init__()
        self.class_names = class_names
        ...
        # Wrap MulticlassAccuracy to log class-wise accuracy for each label
        self.metrics = MetricCollection({
            "acc": ClasswiseWrapper(
                MulticlassAccuracy(num_classes=num_classes, average="none"),
                labels=self.class_names
            )
        })

    def validation_step(self, batch, batch_idx):
        imgs, targets = batch
        preds = self(imgs)
        # Update wrapped metrics; don’t log yet
        self.metrics.update(preds, targets)
        return {"preds": preds, "targets": targets}

    def on_validation_epoch_end(self):
        # Compute all metrics (this returns a dict of tensors) and log them
        metrics_dict = self.metrics.compute()
        self.log_dict(metrics_dict)
        # Reset internal states so next epoch starts fresh
        self.metrics.reset()
```

**Why not call `self.log_dict(self.metrics)` directly?**
Lightning’s `.log_dict()` expects scalar (0-D) tensors or Python floats per key. However, `ClasswiseWrapper(MulticlassAccuracy(...), ...)` returns a 1-D tensor containing one accuracy value per class. By calling:

```python
# ✗ Avoid doing this:
self.log_dict(self.metrics)
```

you’ll run into a shape mismatch whenever Lightning tries to interpret those non-scalar values. Instead, call:

1. **`self.metrics.update(preds, targets)`** during each step to accumulate predictions and targets internally.
2. **`self.log_dict(self.metrics.compute())`** once per epoch to convert everything into a plain `Dict[str, Tensor]` of scalars, which Lightning can handle correctly.
3. **`self.metrics.reset()`** (implicitly or explicitly) to clear out the state before the next epoch.

By adding the missing import and following this pattern—`update() → compute() → log_dict()`—you’ll eliminate the NameError and ensure Lightning logs your class-wise metrics properly.

Issue #716, opened 2025-06-09
incorrect indices used in tiled inference


naomi-simumba:
**Describe the issue**
incorrect indices used in tiled inference

PR #717, opened 2025-06-09
fix indices

naomi-simumba:
issue https://github.com/IBM/terratorch/issues/716


Joao-L-S-Almeida:
Approved, @naomi-simumba . Thanks.

PR #727, opened 2025-06-16
[WIP] Update requirements_test.txt

romeokienzler:


romeokienzler:
requirements_test.txt not used anymore

lhackel-tub:
In the guide for developers (in the readme), it still says to use `pip install -r requirements_test.txt`. What should be used instead /  what is the process for development?

lhackel-tub:
Never mind, I figured it out. One can use poetry install. However, there is an issue with the python version (should be `requires-python = ">=3.10,<4"` instead of `requires-python = ">=3.10"`. We should probably change it in a different PR.

PR #728, opened 2025-06-16
[WiP] Confidence Assement Enhanced Semantic Segmentation

Joao-L-S-Almeida:
Some terms were implemented, but some others point still need be clarified by the authors.


Joao-L-S-Almeida:
No reply from the original authors. No one cares about it.

PR #729, opened 2025-06-17
tiled inference on validation

naomi-simumba:
provide option to do tiled inference during validation


Joao-L-S-Almeida:
HI, @naomi-simumba I agree with the modifications, but my only doubt is if it makes sense to use the tiled inference during the validation, since it affects the quality of the evaluation. Could it negatively impact the fine-tuning ?

Joao-L-S-Almeida (dismissed):

Joao-L-S-Almeida:
@naomi-simumba Do you want to proceed with the merging or prefer to update something before it ?

naomi-simumba:
@Joao-L-S-Almeida ,
I am ready to merge now.

There is another issue when using `DOFA + tiled_inference` where the difference in image sizes between train and val causes an error during the sanity check. This would address that.

Joao-L-S-Almeida (approved):

PR #738, opened 2025-06-23
Clay v1.5 preview

romeokienzler:


requirements_test.txt
Joao-L-S-Almeida, line 1: No more necessary ?

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Approved. Just some comments:
* I restored the 3.13 and for some reason it seems working now.
* Are we going to keep both Clay 1.0 and Clay 1.5 ?

Joao-L-S-Almeida (approved):

PR #753, opened 2025-07-02
Squeezing ground truth with innapropriate formats

Joao-L-S-Almeida:
Trying to adjust ground truth samples when they have extra dimensions. Some datasets, as those created for GeoBench have produced extra unitary dimensions, which could require changes in the configuration. 
```
RuntimeError: only batches of spatial targets supported (3D tensors) but got targets of size: : [64, 1, 224, 224]
```

### DIFF
--- a/terratorch/tasks/segmentation_tasks.py
+++ b/terratorch/tasks/segmentation_tasks.py
@@ -1,8 +1,9 @@
-import warnings
-from typing import Any
-from functools import partial
-import os
 import logging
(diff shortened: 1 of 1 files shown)


Joao-L-S-Almeida:
It tries to solve an issue pointed by @leotizzei when testing `iterate`. This is still in testing mode, @romeokienzler.

romeokienzler (approved):

PR #769, opened 2025-07-10
Label grep still needs  be informed in the prediction mode 

Joao-L-S-Almeida:
A quick fix to avoid issues in prediction mode.

### DIFF
--- a/terratorch/datamodules/generic_pixel_wise_data_module.py
+++ b/terratorch/datamodules/generic_pixel_wise_data_module.py
@@ -3,8 +3,9 @@
 """
 This module contains generic data modules for instantiation at runtime.
 """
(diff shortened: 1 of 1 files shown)


blumenstiel:
@Joao-L-S-Almeida For general understanding: Which issue? Prediction mode should not need a `label_grep`, but we should add it as it is used in all other generic datasets.

blumenstiel (approved):

Joao-L-S-Almeida:
@blumenstiel It was a quick fix, but I believe this was replaced with other PR.

Joao-L-S-Almeida:
Probably a solution as that seen in #770 can be extended to solve it for all the tasks.

PR #776, opened 2025-07-23
Update/release scripts

Joao-L-S-Almeida:
Avoiding some dependencies be pinned in order to allow broader usage of terratorch.

### DIFF
--- a/release/pin_dependencies.py
+++ b/release/pin_dependencies.py
@@ -9,7 +9,7 @@ exclude_list = ["torchvision", "torchmetrics"]
 
 
 def get_pip_list():
-    result = subprocess.run(["pip", "list"], capture_output=True, text=True, check=False)
+    result = subprocess.run(["pip", "list"], capture_output=True, text=True, check=True)
     return result.stdout
 
 


release/pin_dependencies.py
```diff
@@ -1,11 +1,15 @@
...
+
+import toml
+
+# These packages must be left free in order to be adapted according to torch
+exclude_list = ["torchvision", "torchmetrics"]
 
 
 def get_pip_list():
-    result = subprocess.run(["pip", "list"], capture_output=True, text=True)
+    result = subprocess.run(["pip", "list"], capture_output=True, text=True, check=False)
```
blumenstiel: Why would we skip this check? Did you have any errors?

Joao-L-S-Almeida: Strange, I don't modified it. Would it be a ruff change ?

Joao-L-S-Almeida: Yes, it's ruff.

blumenstiel: Thanks for checking!

PR #779, opened 2025-07-30
Trying to avoid  the device be returned as negative

Joao-L-S-Almeida:
It probably indicates that the device is a CPU, but the [negative values ](https://github.com/IBM/terratorch/actions/runs/16624295737/job/47036369501) produce an error when trying to assign devices.

### DIFF
--- a/terratorch/tasks/tiled_inference.py
+++ b/terratorch/tasks/tiled_inference.py
@@ -306,6 +306,11 @@ def tiled_inference(
     original_device = input_batch.get_device()
     input_batch = input_batch.cpu()
 
+    # It handles cases in which the device is
+    # returned as a negative number
+    if original_device < 0:
+        original_device = "cpu"
+
     input_batch_size = input_batch.shape[0]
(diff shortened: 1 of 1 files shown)


terratorch/tasks/tiled_inference.py
```diff
@@ -306,6 +306,11 @@ def tensor_reshape(t):
     original_device = input_batch.get_device()
```
blumenstiel: Whats the differnece between `device` and `original_device`? My understanding is that there both the same. So maybe we should replace `original_device` with `device`? `.device` should work better than `.get_device()` as the function is suppost to return `-1` for CPU. `.device` returns `cpu`.

blumenstiel (changes requested):

terratorch/tasks/tiled_inference.py
Joao-L-S-Almeida: I agree.

blumenstiel: Thx!

PR #789, opened 2025-08-07
Fix metrics

blumenstiel:
Metrics update failed when ignore_index was not set, now flattening the values.


blumenstiel:
See issue: https://github.com/IBM/terramind/issues/17

tests/test_finetune.py
Joao-L-S-Almeida, line 1: Do you consider it isn't necessary to test for the other backbones ?

blumenstiel: That was a mistake, sorry. We should test all of them.

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Thanks, @blumenstiel. Approved.

PR #793, opened 2025-08-10
Trying to pin `transformers`, since something has apparently broken after `4.55.0`

Joao-L-S-Almeida:


tests/test_backbones.py
```diff
@@ -16,6 +17,14 @@
 IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS", "false") == "true"
 
 
+@pytest.fixture(autouse=True)
+def setup_and_cleanup(model_name):
+    hub_dir = os.getenv("HOME") + "/.cache/huggingface/hub"
```
blumenstiel, line 22: We should not download anything from the hub and therefore it should not be required to delete this folder.  Otherwise our stats on HF are broken. Always use `pretrained=False`, which should be the default everywhere.

PR #801, opened 2025-08-14
Fix docs workflow

blumenstiel:
Install timm in docs workflow

Followup from https://github.com/IBM/terratorch/pull/794

Fixing error `ModuleNotFoundError: No module named 'timm'`, see https://github.com/IBM/terratorch/actions/runs/16960714093/job/48072625145#step:4:701

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-          pip install mkdocs-material mike  mkdocstrings[python] mkdocs-git-revision-date-localized-plugin
+          pip install mkdocs-material mike mkdocstrings[python] mkdocs-git-revision-date-localized-plugin timm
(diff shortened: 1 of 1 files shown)


blumenstiel:
Alternatively we can remove this part from the docs. It's not strictly needed.
```
### Timm

You can use any model from `timm` as a backbone. 

!!! tip
    List all available models with `timm.list_models` or filter by name using wildcards:
    
    ```python
    import timm
    timm.list_models('vit*')
    ```

::: timm.list_models
    options:
        heading_level: 4
        show_source: false
```

Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
@blumenstiel  Removing that timm call solved the issue, but now it gives: https://github.com/IBM/terratorch/actions/runs/16963633939/job/48082091365#step:4:754
The PR #802 tries to solve it. It isn't ideal, but we will check if this is the way.

PR #810, opened 2025-08-19
Feature/embedding-generation

Isabelle-Wittmann:
1) Updates TemporalWrapper:
- Extended to Dict Input (SSL4EO/ TerraMind)
- Extended to Dict Output (Terramind Dict Output)
- Replaced For-Loop over timestamps through tensor operations on flattened BxT dim
- Generalized to embedding workflow such that 4d (ViT backbone) outputs are also handled, by unsqueezing to 5d - such that pooling methods can be applied to both 4d and 5d outputs
- Restructured by introducing seperate ‘pool_temporal’ and ‘reshape_5d’ methods
- Input features ‘concat’ and ’n_timestamps’ deprecated; concat introduced as ‘pooling’ option now

2) Introduction Embedding GenerationTask:
- Task to be run with lightning predict workflow, runs inference once over dataset and saves embeddings as parquet or tiff files
- Supports saving of intermediate layer outputs, pooling operations on top (Mean, max, min, cls), and wrapping the model in temporal wrapper
- Sample config for TerraMind embeddings and initial documentation provided 
- To be added: Metadata saving, sample configs for different backbones and data, Tiled Inference Workflows


terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+        Returns:
+            tuple[list[torch.Tensor], list[int]]: Tuple containing list of embeddings and corresponding layer indices.
+        """
+        try:
+            outputs = self.model(input)
+        except Exception as e:
+            raise RuntimeError(f"Model inference failed: {e}")
+        
+        if not isinstance(outputs, list):
+            logging.warning("Model output is not a list; assuming single output layer.")
```
blumenstiel: It's better to use `warnings.warn` because it is only raised once while logging.warning would be logged in every forward pass.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+    Task that runs inference over model backbone to generate and save embeddings.
+    """
+
+    def __init__(
+            self,
+            model: str,
+            model_args: dict = None,
+            output_dir: str = "embeddings",
+            image_key: str = "image",
+            filename_key: str = "filename",
```
blumenstiel: Can we just assume `image` and `filename` or is it possible that there are other names? In all other tasks it's fixed and albumentations is also using these specific key names.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+            raise RuntimeError(f"Model inference failed: {e}")
+        
+        if not isinstance(outputs, list):
+            logging.warning("Model output is not a list; assuming single output layer.")
+            outputs = [outputs]
+
+        n = len(outputs)
+        layers = [l if l >= 0 else n + l for l in layers]
+        valid_layers = [l for l in layers if 0 <= l < n]
+        embeddings = [outputs[l] for l in valid_layers]
```
blumenstiel: I feel like we can just assume that a user defines the correct layers. If not, we could also raise an error instead of fixing it?

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+
+        return embeddings, valid_layers
+    
+    @torch.no_grad()
+    def predict_step(self, batch: dict) -> None:
+        image_key = self.hparams.image_key
+        filename_key = self.hparams.filename_key
```
blumenstiel: See above. I would just use `'image'` if possible.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+            if filenames is None:
+                raise KeyError(f"Key '{filename_key}' not found in input dictionary.")
+            if 'metadata' in batch:
+                metadata = self.pull_metadata(batch['metadata'])
+            else:   
+                metadata = self.pull_metadata(batch)
+
+        embeddings, layers = self.get_embeddings(x, self.hparams.layers)
+
+        fmt = self.hparams.output_format.lower()
```
blumenstiel: Personal view: Should we avoild using `self.hparams` and just save the values as attrs, e.g. `self.output_format`? I guess this is normal practice, as I don't see using `self.hparams` often in other projects.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+            filenames: list[str],
+            metadata: dict,
+            dir_path: Path,
+            fmt: str
+    ) -> None:  
+        """Write a batch (and optional timesteps) to GeoTIFF/GeoParquet."""
+        dir_path.mkdir(parents=True, exist_ok=True)
+
+        B = len(filenames)
+        T = len(filenames[0]) if isinstance(filenames[0], (list, tuple, np.ndarray)) else None
```
blumenstiel: What if the input files include multiple time steps? E.g. in zarr files.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+    ) -> torch.Tensor:
+        """Apply pooling to embeddings."""
+        if pooling in (None, "None", "keep"):
+            return embedding
+
+        if pooling.startswith("vit_"):
+            if embedding.dim() != 2:
+                raise ValueError(f"Expected 2D embedding for ViT pooling, got {embedding.dim()}D.")
+            if has_cls is None:
+                has_cls = embedding.shape[0] % 2 == 1
```
blumenstiel, line 285: Pretty hard assumption as it is not correct for 25% of the cases (if h and w are odd). Let's raise at least a warning. Ideally we could get it from attributes from the model.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+
+        super().__init__()
+        self.save_hyperparameters()
+
+        self._warned_keys: set[str] = set() 
+
+    def _warn_once(self, key: str, msg: str) -> None:
+        if key not in self._warned_keys:
+            logging.warning(msg)
+            self._warned_keys.add(key)
```
blumenstiel: I think you can just use `warnings.warn`. That should do the same. But it does not look as nice as `logging`.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+        dir_path: Path
+    ) -> None:
+        """Write a single sample to GeoParquet."""
+        filename = Path(filename).stem  
+        out_path = dir_path / f"{Path(filename)}_embedding.parquet"
+        arr = embedding.detach().cpu().numpy()
+
+        df = gpd.GeoDataFrame({
+            'embedding': arr.tolist(),}
+        )
```
blumenstiel: Is the metadata missing?

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,331 @@
...
+            arr = arr.reshape(arr.shape[1], sqrt_size, sqrt_size)
+                
+        with rasterio.open( 
+            out_path,
+            "w",
+            driver="GTiff",
+            height=arr.shape[1],
+            width=arr.shape[2],
+            count=arr.shape[0],
+            dtype=arr.dtype
```
blumenstiel, line 342: Can we add the metadata (location, crs), so that the embeddings can be visualized in QGIS etc? Would be a cool feature. We can also add other metadata as tags (e.g. time etc.)

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: filename can have a different meaning here, in the case of zarr files (SSL4EO) extracts individual file_ids instead of zarr file names as in other TT tasks. Will consider with you other feedback to update logic

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: This is already meant as a backup, in case has_cls is not set and can always be overwritten by has_cls setting in config. Agree, from model attributes would be nicer, will check, and raise a warning

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: Yes - metadata to be added.

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: Yes - metadata to be added.

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: We still want the logic of wrapping negative indices for unified naming convention of the subfolders, but changing the valid_layer logic to raising an error

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: _Updated_

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: _Updated_

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: _Updated_

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: _Updated_

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: _Updated_

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: _Updated for 'image'_

terratorch/tasks/embedding_generation.py
blumenstiel: True, I missed that. We can call it `embed_file_key` or something similar?

terratorch/tasks/embedding_generation.py
blumenstiel: That makes sense!

Joao-L-S-Almeida:
@Isabelle-Wittmann @blumenstiel Seemingly the tests have been interrupted due to memory overflow. I replicated them locally and they have used ~ 19 GB of memory, which surpasses the current GH container limitations. Let me try to reduce the tests load. 
Otherwise we could need to upgrade our GH account to use more resources, @romeokienzler.

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: Renamed and included an dimension check based on input data shape

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: See above (embed_file_key)

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,363 @@
...
+        elif pooling == "vit_min":
+            return embedding.min(dim=0).values
+        elif pooling == "vit_cls":
+            return embedding[0, :]
+        elif pooling == "cnn_mean":
+            return embedding.mean(dim=(1, 2))
+        elif pooling == "cnn_max":
+            return embedding.max(dim=(1, 2)).values
+        elif pooling == "cnn_min":
+            return embedding.min(dim=(1, 2)).values
```
blumenstiel, line 307: Does it make sense to use the `ReshapeTokensToImage` neck for ViT outputs? This way we get the same output shape as CNN models and we only need one pooling strategy?

blumenstiel: I also fixed some issues with the neck here: https://github.com/IBM/terratorch/pull/825. So in a next step (after this and the other pr are merged), we could also remove the has_cls param as the neck handles it.

terratorch/tasks/embedding_generation.py
```diff
@@ -0,0 +1,363 @@
...
+        if arr.ndim == 1:
+            arr = arr.reshape(-1, 1, 1)
+        elif arr.ndim == 2: 
+            n_tokens, dim = arr.shape
+            if self.has_cls is True or (self.has_cls is None and n_tokens % 2 == 1):
+                arr = arr[1:, :]
+                n_tokens -= 1
+            s = int(np.sqrt(n_tokens))
+            if s * s != n_tokens:
+                raise ValueError(f"Cannot reshape {n_tokens} tokens into {s}x{s} grid.")
```
blumenstiel, line 332: See above, I think it would be best if we use the TT neck for the reshaping in general.

blumenstiel (dismissed):
Looks good to me!

blumenstiel (dismissed):

blumenstiel (approved):

blumenstiel (approved):

PR #812, opened 2025-08-19
Making Galileo optional

Joao-L-S-Almeida:
As Galileo will be incorporated to TerraTorch, this operation will guarantee the current implementation is tested and avoid issues when installing via PyPI. 
This changes were took since we don't have a defined timeline for the inclusion of Galileo as part of TerraTorch and it can happen just after the release of 1.1

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,7 +47,9 @@ jobs:
           #python -m pip install --upgrade pip
           pip install pip==24.2
           pip --version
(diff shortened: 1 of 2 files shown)


blumenstiel:
@Joao-L-S-Almeida Looks good! Did you already changed the logging messages in a pr before?

Joao-L-S-Almeida:
@blumenstiel Done in the last commit.

PR #822, opened 2025-08-22
Changing documentation branch

Joao-L-S-Almeida:


Joao-L-S-Almeida:
@blumenstiel Please, ignore it. I removed old branches which were blocking the creation of `docs`. The documentation pipeline is now working with the current configs.

PR #825, opened 2025-08-25
Fix ReshapeTokensToImage

blumenstiel:
ReshapeTokensToImage did not infer the grid size correctly. Added the possibility to pass kwargs to a neck. Passing input_size to ReshapeTokensToImage to get patch_size and then the height.

Fall back to remove_cls_token=False, with first error.

### DIFF
--- a/terratorch/models/necks.py
+++ b/terratorch/models/necks.py
@@ -1,5 +1,6 @@
 import math
 import pdb
+import warnings
 from abc import ABC, abstractmethod
 from collections import OrderedDict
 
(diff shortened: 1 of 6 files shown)


Joao-L-S-Almeida:
@blumenstiel It seems that the problem was really the memory exhaustion after After commenting part of the tests related to the temporal wrapper. The tests using clay and swin are apparently to heavyweight (memory uses ~ 12 Gb just for them), so I need to ask if these tests are really indispensable or if we could to  reduce their requirements.

Joao-L-S-Almeida:
It looks like to explicitly deleting the objects before recreating them has produced some relief in the memory and we no longer need to remove those tests. It's strange that the garbage collector wasn't effectively doing that. If you agree, @blumenstiel, we can proceed with the review and merging.

terratorch/models/necks.py
Joao-L-S-Almeida, line 1: This change replacing a Sequential to a ModulesList was just to support extra kwargs, right ?

terratorch/tasks/base_task.py
Joao-L-S-Almeida, line 1: Is it for guaranteeing backcompatibility ?

terratorch/models/necks.py
blumenstiel: Yes. Sequential only take pos args. I expect in the future that we might have multiple different kwargs depending on the necks, so I would not want use pos args.

terratorch/tasks/base_task.py
blumenstiel: Yes exactly, so we can still load old fine-tuned models with a newer TT version.

PR #833, opened 2025-09-02
add TerraMind embedding generation example notebook

GieziJo:
Created an example notebook to generate TerraMind embeddings based on @Isabelle-Wittmann's implementation:
https://github.com/IBM/terratorch/blob/main/examples/confs/embedding_generation/terramind_embeddings.yaml
https://github.com/IBM/terratorch/blob/main/terratorch/tasks/embedding_generation.py

The notebook loads the example `S2L2A/Santiago.tif` file from huggingface, then selects a single 16x16, RBG patch from that image, and outputs the embedded vector of dimension `768`.

The notebook can be found here:
https://github.com/GieziJo/terratorch/blob/TerraMindEmbeddingGenerationNotebook/examples/notebooks/TerraMindEmbedding.ipynb

### DIFF
--- a/examples/notebooks/TerraMindEmbedding.ipynb
+++ b/examples/notebooks/TerraMindEmbedding.ipynb
@@ -1,0 +1,264 @@
+{
+ "cells": [
+  {
+   "metadata": {},
+   "cell_type": "markdown",
+   "source": [
+    "# Embedding Generation\n",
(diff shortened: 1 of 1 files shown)


Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Thanks, @GieziJo. @Isabelle-Wittmann, do you think there is some post-processing stage which would be valid to do for these embeddings ?

PR #840, opened 2025-09-05
improve verbosity on tests

romeokienzler:


Joao-L-S-Almeida:
I didn't see much difference in the logs. Is it already activated ?

PR #843, opened 2025-09-05
Add/temporalwrapper doc

Isabelle-Wittmann:
Add temporal wrapper documentation and sample notebook.
Integrated that out_channels argument from the wrapped backbone is exposed such that TemporalWrapper backbones can be used in TerraTorch EncoderDecoder pipelines.


docs/tutorials/temporal_wrapper.md
blumenstiel, line 1: Should we move this to `docs/guide`? You also have to add it in `mkdocs.yml`, I would add it under `Guide -> Models`.

blumenstiel: Can you add a bit more information on how to use it in a config/python code. E.g. like here for Prithvi: https://ibm.github.io/terratorch/stable/guide/prithvi_eo/

blumenstiel:
@Isabelle-Wittmann Our docs are deployed from the branch `docs`. So we need to merge `main` into `docs` after this pr. `docs` is not protected, which makes it easier to update the docs for us because you don't need pr reviews, you can also do it from the docs directly.

docs/tutorials/temporal_wrapper.md
Isabelle-Wittmann: Moved and extended, Cli/config is currently not possible as TemporalWrapper is not a registered backbone (yet)

blumenstiel: Thanks! I add it quickly then

terratorch/models/utils.py
```diff
@@ -111,7 +111,22 @@ def __init__(self, encoder: nn.Module, pooling: str = 'mean', concat: bool = Non
                 self.reverse_permute_op[p] = i
         else:
             self.reverse_permute_op = None
+
+        if hasattr(encoder, "out_channels"):
```
blumenstiel, line 120: Do we have to raise a error or at least a warning if encoder does not have out_channels? Because that is expected from each backbone in TT. Maybe we can add a out_channels param as a fallback for custom backbones.

blumenstiel:
General question: Should we move the class to it's own py file? I feel like it could become a bigger feature in the future and it would be good if we don't break imports etc. for backwards compatibility etc.

blumenstiel:
@Isabelle-Wittmann The temp wrapper was already supported in configs etc. I updated the code to fit your changes and added it to the docs. Can you review my changes? I also left some new comments

Isabelle-Wittmann:
Great thanks, all looks good! @blumenstiel

PR #851, opened 2025-09-08
[WiP] Workflow for coverage

Joao-L-S-Almeida:


Joao-L-S-Almeida:
@romeokienzler The goal was to generate the coverage-badge and open a PR to merge it with main, but that is failing. 
I also tried to use to upload an artifact, but that "isn't found" just after the upload.

PR #855, opened 2025-09-10
Adding `dask`

Joao-L-S-Almeida:
It looks like `dask` must be mandatory to use the package. 
`diffusers` is already installed by default.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ dependencies = [
   "hdf5plugin",
   "h5netcdf",
   "pytest-sugar",
+  "dask",
 ]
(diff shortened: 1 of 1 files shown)


blumenstiel:
@Joao-L-S-Almeida To use which package? Most package like zarr work with and without dask.

Why are we installing `diffusers` by default? I guess you only need it for the TerraMind generation models. I would not install it if it is not needed in general.

blumenstiel:
Did not see it was closed:)

Joao-L-S-Almeida:
No problem. The package was terratorch itself.

PR #856, opened 2025-09-10
Adding TM tokenizers as backbones

blumenstiel:
Tokenizers were only registered in the FULL_MODEL_REGISTRY sofar. Adding them as backbones as well.


blumenstiel:
@Isabelle-Wittmann I understand the idea but I would keep it this way. Tensor inputs should be the default, dict only if it does not work otherwise. You can use `concat_bands: True` in the general multimodal data module to switch from dict data to tensor data if you only use a single modality.

Just another idea: We can also add a multimodal wrapper similar to the temporal one for late fusion models.

PR #857, opened 2025-09-10
Adding missing dependencies `dask` and `tokenizers`.

Joao-L-S-Almeida:
The package `diffusers` is already included among the mandatory dependencies and don't need to be in `test`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ dependencies = [
   "h5netcdf",
   "pytest-sugar",
   "dask",
+  "tokenizers"
 ]
 
 


blumenstiel:
@Joao-L-S-Almeida To use which package? Most package like zarr work with and without dask.

Why should we install diffusers and tokenizers by default? I guess you only need it for the TerraMind generation models. I would not install it if it is not needed in general.

Joao-L-S-Almeida:
@blumenstiel This is my experience when installing `main` in a new fresh environment (`pip install git+https://github.com/IBM/terratorch.git@main#egg=terratorch[peft]`) and running the unit tests, which will fail with these missing dependencies. 
 If there is any specificity in these tests that aren't applied for general use cases, so let's make it clearer.

blumenstiel:
I understand. But then we should rather skip these tests that require diffusers/tokenizers when the packages are not installed instead of installing them for everyone. I guess 99% of the users are not using them.

blumenstiel:
@Joao-L-S-Almeida Can you check if the problem is solved in https://github.com/IBM/terratorch/pull/863?

Joao-L-S-Almeida:
@blumenstiel Approved. Thanks.

PR #883, opened 2025-09-22
Update timm import

blumenstiel:
Package timm==1.0.17 leads to following warning:
```text
FutureWarning: Importing from timm.models.layers is deprecated, please import via timm.layers
  warnings.warn(f"Importing from {__name__} is deprecated, please import via timm.layers", FutureWarning)
```

Updating import. This warning will still occur because of old imports in Sunya which needs to be updated as well.


Joao-L-S-Almeida (approved):

blumenstiel:
@Joao-L-S-Almeida can you update it in Surya as well so we get rid of the warning for 1.1?

PR #893, opened 2025-09-25
Remove pytest and pytest-sugar from pyproject.toml

romeokienzler:
Removed 'pytest' and 'pytest-sugar' from dependencies.


Joao-L-S-Almeida:
Lgtm.

Issue #880, opened 2025-09-19
Hausdorff (DT) loss (differentiable approximation). Maximizes worst-case boundary errors; sharp edges


Joao-L-S-Almeida:

blumenstiel:
Tested the ER version in kornia and a adaptation of the DT version from monai, see #897. Both did not train on a testset, so I removed them again.

PR #897, opened 2025-09-26
Add boundary-loss

blumenstiel:
Added support for combined losses for segmentation task.

You can define losses in 3 ways:
```
# single loss
    loss: ce 

# equal weights
    loss: 
      - ce
      - dice

# weighted losses
    loss: 
      ce: 0.3
      dice: 0.7
```
Add new loss `lovasz`. Tested two implementations of the `hausdorff` loss but it did not work in tests, so it was removed.
Added new metric: `Boundary_mIoU`.

Other things:
- Removed GeoBench2 dependency from tests
- fix some points in the generic multimodal data module
- added tests

### DIFF
--- a/terratorch/datamodules/__init__.py
+++ b/terratorch/datamodules/__init__.py
@@ -62,9 +62,6 @@ try:
 except ImportError:
     logging.getLogger("terratorch").debug("geobench_v2 not installed")
     geobench_v2_present = False
-except NameError:
-    logging.getLogger("terratorch").debug("geobench_v2 not installed")
-    geobench_v2_present = False
 
 # miscellaneous datamodules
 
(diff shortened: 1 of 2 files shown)


Joao-L-S-Almeida:
@blumenstiel Is this PR totally or partially intersecting with #880 and #908 ?

blumenstiel:
@Joao-L-S-Almeida Yes, both losses listed in the issues are added. I did not use the Haushoff DT loss to avoild further dependencies. The ER version is already included in Kornia which is required by torchgeo.

blumenstiel:
Tested the new losses on Sen1Floods11. Lovasz performed well, similar to dice. Converged a bit faster in the beginning of the training. Hausdorff did not train at all, while the loss is pretty close to 0 the whole time. I could not find an issue in the code, the wrapper provides the values in the correct shape to kornia. Should we remove it again?

PR #902, opened 2025-09-28
Fix GitHub license detection

adamjstewart:
At the moment, GitHub is unable to detect the license of this project. This is because `LICENSE` does not contain any license, only human-readable comments about there being multiple licenses. According to https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#detecting-a-license:

> If your repository is using a license that is listed on the Choose a License website and it's not displaying clearly at the top of the repository page, it may contain multiple licenses or other complexity. To have your license detected, simplify your LICENSE file and note the complexity somewhere else, such as your repository's README file.

The fact that certain files have a different license is already documented in the README, and doesn't need to be documented in the LICENSE file too.

P.S. Happy to discuss how we can contribute some of these MIT licensed files back upstream to TorchGeo! That would also simplify the license situation greatly.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,29 +1,202 @@
-# Project License
 
-## Apache License 2.0
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-All code in this repository is licensed under the Apache License, Version 2.0.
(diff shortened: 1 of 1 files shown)


Joao-L-S-Almeida (approved):

Joao-L-S-Almeida:
Thanks you, @adamjstewart.

PR #916, opened 2025-10-02
Trying to fix number format issue

Joao-L-S-Almeida:


Joao-L-S-Almeida:
`xformers` can't run in CPU. How to disable it to run the tests, @christian-pinto ?

christian-pinto:
I am not an xFormers expert and I have not written this code 😄  but looking at the code it seems to be very easy.
Just set `XFORMERS_AVAILABLE=False` if there is no GPU available and it will revert to the other attention mechanism. This is not just for the tests but it should be done in general, to avoid xformers if there is no GPU. 

If instead you want to only skip the test and leave the code as it is just use a `skipif` decorator for the test function to be skipped if there is not GPU available. However, please note that the test fails on GPU as well though, at least in my environment. For whatever the reason torch defaults to cpu. Perhaps some of the tensors used along the line are on cpu while they should be on gpu if you want to use xformers.

christian-pinto:
In here https://github.com/IBM/terratorch/blob/964e163bf016c4f7c2d45b2c453c026b3e5ae7e7/terratorch/models/backbones/terramind/tokenizer/models/vit_models.py#L32-L37

you could do 

```python
try:
    if torch.cuda.is_available():
        from xformers.ops import memory_efficient_attention, unbind
        XFORMERS_AVAILABLE = True
    else:
        XFORMERS_AVAILABLE = False
except ImportError:
    logging.getLogger('terramind').debug("xFormers not available")
    XFORMERS_AVAILABLE = False
```

or in here:

https://github.com/IBM/terratorch/blob/964e163bf016c4f7c2d45b2c453c026b3e5ae7e7/tests/test_reconstructions.py#L74-L76

you could do:

```python
@pytest.mark.skipif(not torch.cuda.is_available(), reason="skipping test because CUDA is not available")
@pytest.mark.parametrize("model_name", ['terramind_v1_base_generate'])
def test_terramind_generation(model_name):
```
In the first case the correct attention backend is used depending on the device available and the test can be executed on CPU as well. In the second case instead, you always skip the test if you are not on GPU, meaning not testing that functionality on CPU at all. 

I would personally go with the first implementation as it makes sure that the functionality can be tested both on CPU and GPU

Joao-L-S-Almeida:
Solved. @blumenstiel If you agree with these changes, please approve the PR. While we don't have GPUs available on GH, we can always run the integration tests in another machine.

blumenstiel (approved):

PR #923, opened 2025-10-10
Geobench v2 od

paolofraccaro:
Code linked to the detection part of GEo-Bench v2.

### DIFF
--- a/terratorch/models/backbones/terramind/tokenizer/models/uvit.py
+++ b/terratorch/models/backbones/terramind/tokenizer/models/uvit.py
@@ -40,17 +40,6 @@ from diffusers.models.resnet import Downsample2D, Upsample2D
 from diffusers.configuration_utils import ConfigMixin
 from diffusers.models.modeling_utils import ModelMixin
 
-# xFormers imports
-try:
-    if torch.cuda.is_available(): # Only use xformers if GPUs are available
-        from xformers.ops import memory_efficient_attention, unbind
-        XFORMERS_AVAILABLE = True
-    else:
-        XFORMERS_AVAILABLE = False
-except ImportError:
-    logging.getLogger('terramind').debug("xFormers not available")
-    XFORMERS_AVAILABLE = False
-
 
 def modulate(x, shift, scale):
     return x * (1.0 + scale.unsqueeze(1)) + shift.unsqueeze(1)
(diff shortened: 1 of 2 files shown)


terratorch/datasets/generic_pixel_wise_dataset.py
```diff
@@ -289,11 +290,15 @@ def plot(self, sample: dict[str, Tensor], suptitle: str | None = None, show_axes
 
         .. versionadded:: 0.2
         """
+
         image = sample["image"]
         if len(image.shape) == 5:
-            return
+            return 
+        if len(image.shape) == 4:
```
Joao-L-S-Almeida, line 297: Are the dimensions (B,C,T,W,H) ?
T=1 when the len(dims) = 4 ?
Why we get image[:, -1, :, :] before adding the extra dimension ?

paolofraccaro: At this point you do not have the batch. This is taking the latest timestamp in case of temporal data. An improvement would be to allow to select what to plot.

terratorch/tasks/segmentation_tasks.py
```diff
@@ -293,7 +293,15 @@ def configure_metrics(self) -> None:
             )
         else:
             self.test_metrics = nn.ModuleList([metrics.clone(prefix="test/")])
-
+    
+    def reformat_y(self, y):
```
romeokienzler: @paolo-fraccaro 
can we get a comment what this is doing, and why/from where it is supposed to be callsed?

tests/test_temporal_wrapper.py
```diff
@@ -16,6 +16,8 @@ def __init__(self, out_channels=64, in_channels=3):
         super().__init__()
         self.out_channels = out_channels
         self.conv = nn.Conv2d(in_channels, out_channels, kernel_size=3, padding=1)
+
```
romeokienzler: @paolo-fraccaro why are we repeating this line?

terratorch/datasets/generic_pixel_wise_dataset.py
romeokienzler: @Joao-L-S-Almeida can we resolve?

romeokienzler:
@paolo-fraccaro first of all, thanks a lot for the contribution.

I don't see any tests. We introduced a rule that a PR starting from v1.1 is not allowed to decrease our test coverage. Can you please make sure that you cover at lest 80% of your code by tests?

terratorch/datasets/generic_pixel_wise_dataset.py
Joao-L-S-Almeida: Yes, @romeokienzler

terratorch/tasks/segmentation_tasks.py
paolofraccaro: I removed this as there was already a squeeze of they y in main.

tests/test_temporal_wrapper.py
paolofraccaro: typo

PR #924, opened 2025-10-10
Running the tests inside container environments (for local execution)

Joao-L-S-Almeida:
To run the tests inside a container environment, proceed in the following way:

* Install `podman` in your system.
* Building the base images:
```
./run_base_images_build.sh 
```
* Running the tests inside each environment:
```
./run_tests_container.sh 
```

### DIFF
--- a/run_base_images_build.sh
+++ b/run_base_images_build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-PYTHON_VERSIONS=("3.11") # "3.12" "3.13")
+PYTHON_VERSIONS=("3.11" "3.12" "3.13")
 
 for v in ${PYTHON_VERSIONS[@]};
 do


Joao-L-S-Almeida:
To be finished within two weeks or it will be closed.

Joao-L-S-Almeida:
@christian-pinto Maybe you could consider it, together with the `tox` solution seen in #994, as a solution for local reproducibility.

Dockerfile.base
```diff
@@ -0,0 +1,12 @@
+ARG version
+FROM python:$version-trixie
+
+RUN echo "Preparing environment for Python $version"
+RUN apt update
+RUN apt install -y python3-pip bash 
+
+RUN git clone https://github.com/terrastackai/terratorch.git
```
christian-pinto: You can copy the terratorch source code form the local directory, no need for a git checkout.

Dockerfile.pytest
```diff
@@ -0,0 +1,10 @@
+ARG version
+FROM terratorch_python_$version:latest
+
+RUN echo "Preparing environment for Python $version"
+
+COPY pyproject.toml .
+COPY terratorch terratorch
+COPY tests tests
```
christian-pinto, line 15: Isn't terratorch already present as part of the base image you are building this one from?

christian-pinto:
I personally do not see the value of running inside a container if we are adopting tox that will make sure a fresh virtual environment is created for the version of python we are testing for.

christian-pinto:
Apologies, I see you mention for local execution only, yes in that case one might want to create a container with multiple version of python installed. In the case of the GH workflows you'll probably going to get python pre-installed wherever the test is running.

run_base_images_build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+PYTHON_VERSIONS=("3.11" "3.12" "3.13")
+
+for v in ${PYTHON_VERSIONS[@]};
+do
+  echo "Creating base image for Python $v"
+  podman build --build-arg version=$v -f Dockerfile.base -t terratorch_python_$v
```
christian-pinto: Do we have any dependency on podman? I would rather use docker as in the end there's mappings for the docker cli to the podman backend.

run_pytest_images_build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+PYTHON_VERSIONS=("3.11" "3.12" "3.13")
+
+for v in ${PYTHON_VERSIONS[@]};
+do
+  echo "Preparing test environment for Python $v"
+  podman build --build-arg version=$v -f Dockerfile.run -t "pytest_terratorch_$v"
+done
+
```
christian-pinto: You can build an image with all the python versions installed and then run tox inside to execute the tests.

Dockerfile.base
Joao-L-S-Almeida: Yes, in fact. Done in the last commit.

Dockerfile.pytest
Joao-L-S-Almeida: Modified in the last commit.

run_base_images_build.sh
Joao-L-S-Almeida: I usually prefer podman because it is rootless. I understand it's available in all the systems: https://podman.io/docs/installation

Joao-L-S-Almeida: The Dockerfiles should work both for podman and docker, so maybe it's a matter of adding this option in the running scripts.

run_pytest_images_build.sh
Joao-L-S-Almeida: We could try, but the Python images (built on Debian) are distributed with fixed versions and install more than one in there could break or mess something. 
In any case, the cost would be very similar, since the environments would be recreated and the dependencies downloaded again for each Python version.

PR #929, opened 2025-10-13
ScalarRegressionTask

leonardijulia:
This PR adds the new ScalarRegressionTask, enabling image-to-scalar prediction for both single-variable and multi-variable regression, including optional per-variable weighting for losses and metrics. Pixel-wise multivariate regression and its required supporting utilities are not yet implemented and remain as TODO items.

### DIFF
--- a/terratorch/tasks/regression_tasks.py
+++ b/terratorch/tasks/regression_tasks.py
@@ -59,7 +59,7 @@ class WeightedMultivariateLossWrapper(nn.Module):
     
     def forward(self, output: Tensor, target: Tensor,) -> Tensor:
         loss = self.loss_function(output, target)
-        weights = self.weights.view(1, -1, * ([1] * (loss.ndim - 2))) # [1, num_vars, 1, 1] for pixel-wise (TODO: Pixel-wise multivariate)
+        weights = self.weights.view(1, -1, * ([1] * (loss.ndim - 2))) # [1, num_vars, 1, 1] for pixel-wise
                 
         weighted_loss = weights * loss
         
@@ -67,7 +67,7 @@ class WeightedMultivariateLossWrapper(nn.Module):
             return weighted_loss # [B, num_vars, H, W] or [B, num_vars]
         
         if self.reduction == "mean": 
-            return weighted_loss.sum() / weights.sum()  # weighted scalar mean for backprop
+            return weighted_loss.sum() / weights.expand_as(loss).sum()  # weighted scalar mean for backprop
         
         msg = "Only 'mean' and None reduction supported"
         raise Exception(msg)


leonardijulia:
@blumenstiel

terratorch/models/encoder_decoder_factory.py
```diff
@@ -111,7 +111,7 @@ def build_model(
             backbone_kwargs (dict, optional) : Arguments to be passed to instantiate the backbone.
             decoder_kwargs (dict, optional) : Arguments to be passed to instantiate the decoder.
             head_kwargs (dict, optional) : Arguments to be passed to the head network. 
-            num_classes (int, optional): Number of classes. None for regression tasks.
+            num_classes (int, optional): Number of classes or regression variables.
```
blumenstiel: Can we rename this one to `num_outputs` or similar?

terratorch/models/heads/__init__.py
```diff
@@ -1,7 +1,7 @@
 # Copyright contributors to the Terratorch project
 
-from terratorch.models.heads.classification_head import ClassificationHead
+from terratorch.models.heads.scalar_head import ScalarHead
 from terratorch.models.heads.regression_head import RegressionHead
 from terratorch.models.heads.segmentation_head import SegmentationHead
```
blumenstiel, line 5: Can you add something like
```
# TODO: Remove in a version v1.3 or later
@deprecated
ClassificationHead = ScalarHead
```
and add it to all? Just for backwards compatibility.

terratorch/models/heads/regression_head.py
```diff
@@ -26,6 +26,7 @@ class RegressionHead(nn.Module):
     def __init__(
         self,
         in_channels: int,
+        num_classes: int = 1,
```
blumenstiel: I would call it `num_outputs` or `num_variables`. Classes is not the best one for regression, right?

terratorch/models/encoder_decoder_factory.py
blumenstiel: Ah wait, for classification it should still be `num_classes`. Maybe a second param and depending on the task either classes or outputs get passed to the head? Or something like
```
num_outputs = num_classes or num_variables
```
before passing it to the head as `num_outputs`. But separate might be better so `SegmentationHead` does not change.

tests/test_prithvi_tasks.py
```diff
@@ -4,7 +4,7 @@
 import torch
 
 from terratorch.models.backbones.prithvi_vit import PRETRAINED_BANDS
-from terratorch.tasks import ClassificationTask, PixelwiseRegressionTask, SemanticSegmentationTask, ReconstructionTask
+from terratorch.tasks import ClassificationTask, PixelwiseRegressionTask, SemanticSegmentationTask, ReconstructionTask, ScalarRegressionTask
```
blumenstiel: Not needed?

tests/test_prithvi_tasks.py
```diff
@@ -168,6 +168,9 @@ def test_create_classification_task_encoder_decoder(backbone, decoder, loss, mod
 
     gc.collect()
 
+
+
+
```
blumenstiel: lot of empty lines

tests/test_tasks.py
```diff
@@ -0,0 +1,59 @@
...
+        class_weights=class_weights
+    )
+    
+    batch = {
+        "image": torch.ones((1, 6, 224, 224)),
+        "mask": torch.randn((1, num_classes))
+    }
+    
+    model_output : ModelOutput = task(batch["image"])
+    y_hat = model_output.output
```
blumenstiel, line 72: maybe add some asserts to check the expected output shape

tests/test_tasks.py
```diff
@@ -0,0 +1,59 @@
...
+    y = batch["mask"]
+    
+    loss = task.training_step(batch, batch_idx=0)
+    
+    # Metrics sanity check
+    task.train_metrics.update(y_hat, y)
+    computed_metrics = task.train_metrics.compute()
+    assert isinstance(computed_metrics, dict), "Metrics did not return a dict"
+    print(f"\nMetrics: {computed_metrics}")
```
blumenstiel: no prints in tests

terratorch/tasks/regression_tasks.py
```diff
@@ -44,6 +44,31 @@ def forward(self, output: Tensor, target: Tensor) -> Tensor:
...
+        self.loss_function = loss_function
+        self.register_buffer("weights", weights)
+        self.reduction = reduction
+    
+    def forward(self, output: Tensor, target: Tensor,) -> Tensor:
+        loss = self.loss_function(output, target)
+        if self.weights.ndim == 1:
+            weights = self.weights.view(1, -1, *([1] * (loss.ndim - 2))) # [1, num_vars, 1, 1]
+        else:
+            weights = self.weights
```
blumenstiel: my assumption would be that weights are always ndim == 1? 
So maybe just add a check in init and raise an error otherwise. Or is there a setting where weights would have more than one dim?

terratorch/tasks/regression_tasks.py
```diff
@@ -91,21 +116,26 @@ class IgnoreIndexMetricWrapper(WrapperMetric):
...
+                items_to_take[torch.isnan(target)] = False  # Filter NaN values as well
+                target = target[items_to_take]
+                preds = preds[items_to_take]
+            else:
+                preds = preds.flatten()
+                target = target.flatten()
+            return self.metric.update(preds, target)
+        
+        elif self.num_outputs > 1:
+            return self.metric.update(preds, target)
```
blumenstiel, line 142: Add warning or NotImplementedError if ignore_index is not None.

terratorch/tasks/regression_tasks.py
```diff
@@ -414,3 +482,314 @@ def model_forward(x, **kwargs):
...
+            if self.num_classes > 1:
+                if self.class_weights is not None:
+                    check_weights_classes(self.class_weights, self.num_classes)
+                    weighted_metrics = {
+                        name: WeightedMetricWrapper(metric, self.class_weights) for name, metric in metrics.items() if name != "R2_Score"
+                        }
+                    weighted_metrics["R2_Score"] = metrics["R2_Score"]
+                    metrics = weighted_metrics
+                    
+                per_class_metrics = {name: ClasswiseWrapper(metric, prefix="_") for name, metric in metrics.items() if name != "R2_Score"}
```
blumenstiel: Maybe I am missing something by why are you applying the WeightedMetricWrapper if you then create classwise metrics? Either you report classwise metrics, and you don't need any weights, or only weighted ones, or ideally both together.

blumenstiel: Why does it not work for R2?

terratorch/tasks/regression_tasks.py
```diff
@@ -414,3 +482,314 @@ def model_forward(x, **kwargs):
...
+        """
+        x = batch["image"]
+        file_names = batch["filename"] if "filename" in batch else None
+        other_keys = batch.keys() - {"image", "mask", "filename"}
+        rest = {k: batch[k] for k in other_keys}
+
+        def model_forward(x, **kwargs):
+            return self(x, **kwargs).output
+
+        if self.tiled_inference_parameters:
```
blumenstiel: tiled inference probably dies not work with a scalar task, so you can also remove it I guess

blumenstiel: yes, see classification task

terratorch/models/heads/regression_head.py
leonardijulia: Yes, I was trying to use the same name across all tasks for simplicity but it might introduce more confusion...

terratorch/models/encoder_decoder_factory.py
leonardijulia: yes, that was exactly my thought process - I will add a num_variables and handle it accordingly

terratorch/tasks/regression_tasks.py
leonardijulia: I wanted to report class-wise metrics that are weighted so they are consistent with the computed per-class loss, but perhaps you're right that raw metrics should be passed without weighting, along with the total weighted loss (i.e., the one used in backprop). As for R2, I fixed it now, and it works - I was using a dummy batch of batch size 1, which resulted in R2 being undefined - to calculate classwise R2, the batch size needs to be > 1

terratorch/tasks/regression_tasks.py
leonardijulia: I was thinking of a spatial weighting, but didn't go through with it - I think it could be a future enhancement, but for now I'll just check in init and accept weights.ndim == 1 only

terratorch/models/encoder_decoder_factory.py
```diff
@@ -53,14 +61,20 @@ def _get_decoder_and_head_kwargs(
...
+    
+    if num_outputs is not None and num_classes is not None:
+        raise ValueError("Only one of `num_outputs` or `num_classes` should be provided.")
+    
+    value = num_outputs if num_outputs is not None else num_classes
+    key = "num_outputs" if num_outputs is not None else "num_classes"
+    if value is not None:
         if decoder_includes_head:
-            decoder_kwargs["num_classes"] = num_classes
+            decoder_kwargs[key] = value
```
blumenstiel: For a decoder that includes a head (which I don't know we have one), I would just keep it as "num_classes" as classification is more common and I would expect than these decoders would fail with nun_outputs. 
`decoder_kwargs["num_classes"] = value`

terratorch/models/encoder_decoder_factory.py
```diff
@@ -53,14 +61,20 @@ def _get_decoder_and_head_kwargs(
         )
         logging.warning(msg)
         decoder_includes_head = DECODER_REGISTRY.find_registry(decoder).includes_head
-    if num_classes is not None:
+    
+    if num_outputs is not None and num_classes is not None:
+        raise ValueError("Only one of `num_outputs` or `num_classes` should be provided.")
```
blumenstiel: I would move that to the top

terratorch/models/encoder_decoder_factory.py
```diff
@@ -53,14 +61,20 @@ def _get_decoder_and_head_kwargs(
         )
         logging.warning(msg)
```
blumenstiel: We can make that nicer by checking `if hasattr(DECODER_REGISTRY.find_class(decoder), "includes_head")` instead of a try except block.

terratorch/models/encoder_decoder_factory.py
```diff
@@ -53,14 +61,20 @@ def _get_decoder_and_head_kwargs(
...
         decoder_includes_head = DECODER_REGISTRY.find_registry(decoder).includes_head
-    if num_classes is not None:
+    
+    if num_outputs is not None and num_classes is not None:
+        raise ValueError("Only one of `num_outputs` or `num_classes` should be provided.")
+    
+    value = num_outputs if num_outputs is not None else num_classes
```
blumenstiel: you can write this line also as `value = num_outputs or num_classes` which looks a bit cooler:) I would maybe not name it `value` as it is not very specific. Maybe just keep calling it `num_outputs` or `num_classes` from here on?

terratorch/models/scalar_output_model.py
```diff
@@ -119,10 +119,12 @@ def forward(self, x: torch.Tensor, **kwargs) -> ModelOutput:
         return ModelOutput(output=mask, auxiliary_heads=aux_outputs)
 
     def _get_head(self, task: str, input_embed_dim: int, head_kwargs: dict):
-        if task == "classification":
-            if "num_classes" not in head_kwargs:
-                msg = "num_classes must be defined for classification task"
+        if task == "classification" or task == "scalar_regression":
+            if "num_outputs" not in head_kwargs and "num_classes" not in head_kwargs:
+                msg = f"Either `num_outputs` or `num_classes` must be defined for task `{task}`."
```
blumenstiel: I don't think it should require `num_outputs` for regression. It can just fall back to 1 if not provided as this is the default case. Maybe we can handle that on EncDecFactory level to cover it for pixel-wise regression as well.

terratorch/tasks/regression_tasks.py
```diff
@@ -436,3 +500,306 @@ def model_forward(x, **kwargs):
...
+    def training_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> Tensor:
+        """Compute the train loss and additional metrics.
+
+        Args:
+            batch: The output of your DataLoader.
+            batch_idx: Integer displaying index of this batch.
+            dataloader_idx: Index of the current dataloader.
+        """
+        x = batch["image"]
+        y = batch["mask"]
```
blumenstiel: Let's not call it `mask` for scalar regression. For classification it is called `label` if I remember it correctly. I assume that would also work for scalar regression?

blumenstiel: Also needs to change in the tests.

terratorch/models/encoder_decoder_factory.py
```diff
@@ -32,35 +32,47 @@ def _get_decoder_and_head_kwargs(
...
-            f"Decoder {decoder} does not have an `includes_head` attribute. Falling back to the value of the registry."
-        )
+    # depending on that, pass num classes/num outputs to either head kwrags or decoder
+    if hasattr(DECODER_REGISTRY.find_class(decoder), "includes_head"):
+        msg = (f"Decoder {decoder} does not have an `includes_head` attribute. Falling back to the value of the registry.")
         logging.warning(msg)
-        decoder_includes_head = DECODER_REGISTRY.find_registry(decoder).includes_head
-    if num_classes is not None:
+    
+    decoder_includes_head = DECODER_REGISTRY.find_registry(decoder).includes_head
```
blumenstiel: I think this should be in the if case and the msg in a else case (if tne attr is not there). I also think logging.debug should be sufficient, no point in warning every user.

Also `decoder_includes_head = False`  must be added in the else case. And you could call it `includes_head` to match with the code above. 

(Not related to your pr but I think it make sense to clean this up, thanks:)

terratorch/models/heads/__init__.py
```diff
@@ -1,7 +1,11 @@
...
 
-__all__ = ["ClassificationHead", "RegressionHead", "SegmentationHead"]
+# TODO: Remove in a version v1.3 or later
+# Deprecated alias for backward compatibility
+ClassificationHead = ScalarHead
```
blumenstiel, line 9:
```
from warnings import deprecated

@deprecated("Use ScalarHead instead.")
ClassificationHead = ScalarHead
````
Can you add a `deprecated` decorator to warn users?

File: terratorch/models/heads/scalar_head.py [hunk 20]
```diff
@@ -19,14 +20,17 @@ def __init__(
...
+            num_classes (int, optional): Number of output classes for classification
             dim_list (list[int] | None, optional):  List with number of dimensions for each Linear
                 layer to be created. Defaults to None.
             dropout (float, optional): Dropout value to apply. Defaults to 0.
             linear_after_pool (bool, optional): Apply pooling first, then apply the linear layer. Defaults to False
         """
         super().__init__()
-        self.num_classes = num_classes
+        if num_outputs is None and num_classes is None:
+            raise ValueError("`num_outputs` or `num_classes` should be provided.")
```
blumenstiel, line 33: I would add raise a `warning.warn` if both are provided (code is using num_outputs).

terratorch/models/heads/scalar_head.py
```diff
@@ -19,14 +20,17 @@ def __init__(
...
+        self.num_outputs = num_outputs or num_classes
```
blumenstiel, line 39: I was just thinking about it and it probably make sense to still save `self.num_classes = num_classes` if num_classes is not None because of backwards compatibility.

blumenstiel: In case someone or some code is calling model.head.num_classes or similar

terratorch/tasks/classification_tasks.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         aux_heads: list[AuxiliaryHead] | None = None,
         aux_loss: dict[str, float] | None = None,
         class_weights: list[float] | None = None,
-        ignore_index: int | None = None,
+        ignore_index: int | None = -100,
```
blumenstiel, line 69: Standard value is `-1` most of the time

leonardijulia: yes, I know, but I wanted to match the value that is the default in torch.nn.functional.py::cross_entropy, which is -100.

blumenstiel:
@leonardijulia Sorry, did not check this pr in a while. Is it ready to review or do you have some open todos?
related to #1029

leonardijulia:
@blumenstiel It's ready for review :) I will check out #1029 tomorrow!

PR #930, opened 2025-10-13
Update vLLM tiled parameters

blumenstiel:
TerraTorch tiled inference allows users to specify only the `crop`. Updated vLLM plugin to support this setting and added a warning for `stride`.


romeokienzler:
tnx @blumenstiel 

- [x] merged

PR #931, opened 2025-10-14
fix/dependencies

blumenstiel:
Moved dependencies for surya to install `terratorch[surya]`  because of errors with numba when using torch.distributed.

Also, removed `mlflow` because noone is really using it outside TT interate (it should be a dependency there). I assume wandb and tensorboard are the standard loggers. 

I could not find any usage of `transformers` and `termcolor` in the code, so I removed it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,8 @@ dependencies = [
   "tifffile",
   "python-box",
   "tqdm",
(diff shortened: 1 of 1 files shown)


romeokienzler:
tnx a lot @blumenstiel

PR #935, opened 2025-10-17
[WIP] Enhance pylint workflow with ImportError check

romeokienzler:
Added a check for invalid ImportError catching in non-__init__.py files.


Joao-L-S-Almeida:
Maybe we could move it to 1.1.2

PR #938, opened 2025-10-17
[WIP] Add GitHub Actions workflow for self-hosted runner

romeokienzler:


Joao-L-S-Almeida:
@romeokienzler In which circumstances the file `ccc.yml` will be executed ? In which case it could fail ?

Joao-L-S-Almeida:
Maybe we could move it to 1.1.2.

PR #957, opened 2025-10-22
Add diff_subset option temporal wrapper, update doc and tutorial notebook

Isabelle-Wittmann:
Include diff_subset option for temporal wrapper, which computes the mean across two temporal subsets (lengths defined by `diff_subsets_lengths` parameter). It requires the input samples to all be equal length (eg. T=5 timesteps), and always applies the same temporal split. Eg. when setting diff_subsets_lengths=[2,3] it splits the input into a 2-sample and 3-sample subset, averages and then computes the difference. Only defined for 2 subsets, in order to compute a meaningful difference. 

(For more than 2 subsets we could consider to introduce a mean_subset option in the future, that computes mean per subsets and then mean across means.)

Updated documentation accordingly, and extended the tutorial notebook. Will update tests depending on [PR #956](https://github.com/IBM/terratorch/pull/956)


docs/guide/temporal_wrapper.md
```diff
@@ -7,8 +7,9 @@ class TemporalWrapper(nn.Module):
     def __init__(
         self,
         backbone: nn.Module,
-        pooling: Literal["keep", "concat", "mean", "max", "diff"] = "mean",
+        pooling: Literal["keep", "concat", "mean", "max", "diff", "diff_subsets"] = "mean",
```
blumenstiel: Why do we need a additional method? Is it now the same as `diff` (assuming the subset lengths 1, 1)?

terratorch/models/utils.py
```diff
@@ -148,6 +162,12 @@ def pool_temporal(self, stacked: torch.Tensor, pooling: str):
             return self.subtract_along_dim1(stacked) # difference between first two timesteps
         elif pooling == "mean":
             return torch.mean(stacked, dim=1) # mean over time
+        elif pooling == "diff_subsets":
+            # Split stacked tensor along time according to subset lengths
+            subset_splits = torch.split(stacked, self.subsets_lengths, dim=1)
```
blumenstiel: What happens if subset_lengths does not work with the input in terms of image shape?
Either we can check and raise a warning or we can also allow `-1` in the first or second position and update it according to the input shape.

terratorch/models/utils.py
```diff
@@ -89,6 +90,11 @@ def __init__(self, encoder: nn.Module, pooling: str = 'mean', concat: bool = Non
             features_permute_op (list): Permutation operation to perform on the features before aggregation.
                 This is in case the features to do not match either 'BCHW' or 'BLC' formats. It is reversed once
                 aggregation has happened.
+            diff_subsets_lengths (list[int]): 
+                Defines the temporal lengths of the two subsets used when pooling='diff_subsets'. 
+                The wrapper first computes the mean across each subset and then returns their difference. 
+                The lengths should sum to the total number of timesteps in the input. 
+                Example: [2, 3] averages the first 2 and last 3 timesteps before differencing.
```
blumenstiel: Just some ideas to generalize the approach: 
Could we have a general parameter `subset_lengths` that averages the inputs of the subsets independent of the method? E.g. we could also do `concat` with subsets `[4, 4, 4, 4]` if you have seasonal images. In forward we first run this `if self.subsets_lengths is not None`. If `padding='diff'` we check if len(subsets) == 2 and raise an error otherwise. We can also allow a `-1` in the subsets to allow for variable input lengths.

blumenstiel (changes requested):
Looks good in general. I think we don't need the extra `diff_subsets` method, would also work with `diff`. Can you add a test to check the subset params?

terratorch/models/utils.py
Isabelle-Wittmann: Good point, I will generalize it to multiple subsets, and to not have separate methods.

terratorch/models/utils.py
Isabelle-Wittmann: I didn't implement the checking whether sum(subset_lengths) equals total timesteps, as we can't check this during the construction (n_timesteps often is None). For both the "normal diff" and the subset method, we start from the beginning and eg. use the first two, five, .. timesteps and ignore the rest. 

How about these defaults

- Default for subset lengths is to use [1,1,1..] for given timesteps, so "standard" pooling, for diff we use the first two samples so [1,1] but warn that others are ignored
- if sum(subset_lengths) < T, we append one subset length with T-sum(subset_lengths) for all remaining timesteps, but warn (Or we append [1,1,1...] for all remaining timesteps? @blumenstiel )
- sum(subset_lengths) > T throws an error
- We can allow `-1 ` for one length, but I would advise to set all dims explicit

Should we make `n_timesteps` a required input argument to check all of this during the construction instead of the error only occurring after starting a training? @blumenstiel

Isabelle-Wittmann: What we might loose a bit is the freedom to throw in a different number of timesteps in into a fine-tuned pipeline, this would only work for [1,1,1...] standard pooling out-of-the-box, because the subsets lengths only work for the given T timesteps

blumenstiel: @Isabelle-Wittmann We can just set the subset_lengths default to None. Only if a user explicitly provides them, we mean aggregate before. diff would still work as before with two timestamps.

If subset_lengths is provided, we can run the sum(subset_lengths) < T check in the forward pass.

Then let's drop the -1 idea, you are right that it's probably better make it explicit.

`n_timesteps` can stay only required for concat. However, if subsets is provided as well we need to check that `n_timesteps == len(subset_lengths)` and otherwise raise an error.

blumenstiel: Same, if `diff and subset_lengths is not None` then `len(subset_lengths) == 2` and otherwise raise an error.

Isabelle-Wittmann:
@blumenstiel Updated: 
- Unified `subset_lengths` param to work with all pooling methods, I just switched to checking sum(`subset_lengths`) = T, and also checking whether two (aggregated) timesteps are provided for 'diff'. Both raise errors now, which I think helps to prevent silent misunderstandings. 
- I added tests for the new features, which I will incorporate into the revised test file in https://github.com/IBM/terratorch/pull/956 - for now I just fixed the indent error in the test file such that the new tests are run.

blumenstiel (approved):
Thanks!

PR #922, opened 2025-10-06
[WiP] Adding downstream tasks for surya

Joao-L-S-Almeida:

PR #967, opened 2025-10-29
Support for inference using Surya downstream models

Joao-L-S-Almeida:
This is a partial PR and advances some of the features which will be completed with PR #922. 
This PR is paving the way to remove submodules and strict dependencies for externally-provided models.

### DIFF
--- a/terratorch/__init__.py
+++ b/terratorch/__init__.py
@@ -1,5 +1,9 @@
 # Copyright contributors to the Terratorch project
 
+# Mute albumentations version info
+import os
+os.environ["NO_ALBUMENTATIONS_UPDATE"] = "True"
+
 import terratorch.models  # noqa: F401
 from terratorch.models.backbones import *  # register models in registries # noqa: F403
 from terratorch.registry import BACKBONE_REGISTRY, DECODER_REGISTRY, MODEL_FACTORY_REGISTRY, FULL_MODEL_REGISTRY  # noqa: F401
--- a/terratorch/cli_tools.py
+++ b/terratorch/cli_tools.py
@@ -64,8 +64,6 @@ from terratorch.tasks import (
 from terratorch.utils import compute_mask_statistics, compute_statistics
 
 logger = logging.getLogger("terratorch")
-# Mute albumentations version info
-logging.getLogger("albumentations.check_version").setLevel(logging.WARNING)
 
 from terratorch.utils import remove_unexpected_prefix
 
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -64,6 +64,8 @@ dependencies = [
   "tqdm",
   "wandb",
   "tensorboard",
+  # It will be added when the package is released
```
romeokienzler: shouldn't it go to a different section, e.g.

[surya]

terratorch/models/backbones/__init__.py
```diff
@@ -14,6 +14,6 @@
 
 try:
     import terratorch.models.backbones.heliofm_register
```
romeokienzler, line 17: this one we are keeping? or should it go to the surya package as well?

terratorch/models/backbones/heliofm_register.py
```diff
@@ -1,9 +1,25 @@
 import logging
+
 import huggingface_hub
-from torch import nn
 import torch
+from terratorch_surya.downstream_examples.ar_segmentation.models import (
```
romeokienzler, line 5: did you test if the code is not breaking in case surya is not present in the venv and you run a different e.g. EO task?

terratorch/models/surya_model_factory.py
```diff
@@ -0,0 +1,67 @@
+import importlib
```
romeokienzler: also please make sure this code doesn't break other tasks in case surya is not installed

pyproject.toml
Joao-L-S-Almeida: That is right.

terratorch/models/backbones/__init__.py
Joao-L-S-Almeida: The register will still be in terratorch, but inside it the models are invoked from the surya package. If the package isn't installed it won't be imported (it's bypassed) and the general working isn't affected.

terratorch/models/backbones/heliofm_register.py
Joao-L-S-Almeida: It's a good test.

Joao-L-S-Almeida: I made the imports conditional and skipped the related tests when `terratorch_surya` isn't installed.

pyproject.toml
Joao-L-S-Almeida: Done.

terratorch/models/surya_model_factory.py
Joao-L-S-Almeida: I removed it. It'll be part of the PR #922.

romeokienzler (approved):

PR #989, opened 2025-11-04
Geobench from GH

Joao-L-S-Almeida:


pyproject.toml
```diff
@@ -126,7 +126,7 @@ peft = [
 ]
 
 geobenchv2 = [
-  "geobench_v2",
+  "geobench_v2@git+https://github.com/The-AI-Alliance/GEO-Bench-2.git@main",
```
christian-pinto:
```suggestion
  "geobenchv2@git+https://github.com/The-AI-Alliance/GEO-Bench-2.git@main",
```

Otherwise, If I try to install with the optional `geobenchv2` I get the below error

```
  WARNING: Generating metadata for package geobench_v2 produced metadata for project name geobenchv2. Fix your #egg=geobench_v2 fragments.
Discarding git+https://github.com/The-AI-Alliance/GEO-Bench-2.git@main: Requested geobenchv2 from git+https://github.com/The-AI-Alliance/GEO-Bench-2.git@main (from terratorch==1.1->terratorch==1.1) has inconsistent name: expected 'geobench-v2', but metadata has 'geobenchv2'
INFO: pip is looking at multiple versions of terratorch[geobenchv2] to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement geobench-v2 (unavailable) (from terratorch[geobenchv2]) (from versions: none)

```

christian-pinto: Even with that change there is an issue with the dependencies between terratorch and geobenchv2

```
pip install "dist/terratorch-1.1-py3-none-any.whl[geobenchv2]"

...
...
ERROR: Cannot install terratorch and terratorch[geobenchv2]==1.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    terratorch 1.1 depends on segmentation-models-pytorch<=0.4
    terratorch[geobenchv2] 1.1 depends on segmentation-models-pytorch<=0.4
    torchgeo 0.7.2 depends on segmentation-models-pytorch>=0.5
    torchgeo 0.7.1 depends on segmentation-models-pytorch>=0.5
```

romeokienzler (dismissed):

christian-pinto (changes requested):

pyproject.toml
romeokienzler: thanks @christian-pinto I think https://pypi.org/search/?q=geobenchv2 it was not yet released, so it can't work

@paolo-fraccaro r u planning to release today? cc @Joao-L-S-Almeida 

#992 will fix the the deps problem

christian-pinto: If you see my suggestion the problem is in the package name that is not `geobench_v2` but `geobenchv2`. I have suggested a change that I have tested to be working. 

It should be

```
geobenchv2@...
```

This is not related to a geobenchv2 release.

Joao-L-S-Almeida: I believe the PR #995 should solve the conflict with `smp`.

PR #994, opened 2025-11-07
Configuration for `tox`

Joao-L-S-Almeida:
This configuration allows the execution of local tests for multiple versions in environments re-created on demand. 
To run the tests for all the versions, install `tox` (via `pip` or `uv`) and
```
tox run
```
For a specific version:
```
tox run run -e 3.13
```
In case you don't have some Python versions installed, `tox` will skip them.

### DIFF
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -35,6 +35,18 @@ modification be merged to `main`. In this way, when an user wants to modify
     ```
     pytest -s -v tests/
     ```
+* To run the tests for all the versions:
+1. install `tox` (via `pip` or `uv`)
+2. Run `tox`:
+```
+tox run
+```
+3. For a specific version:
+```
+tox run run -e 3.13
+```
+
+In case you don't have some Python versions installed, `tox` will skip them. 
 * If all the tests are passing, you can open a PR to `terratorch:main` describing what you are adding and why
     that is important to be merged. You
     do not need to choose a reviewer, since the maintainers will check the new open PR and request review for it by themselves.  


tox.ini
```diff
@@ -0,0 +1,13 @@
...
+requires =
+    tox >= 4.20
+env_list =
+    3.13
+    3.12
+    3.11
+
+[testenv]
+deps =
+    git+https://github.com/The-AI-Alliance/GEO-Bench-2.git@main
```
christian-pinto: I cannot test this PR because GEO-Bench-2  has conflicting dependencies with TerraTorch. 

I believe we should first implement https://github.com/terrastackai/terratorch/issues/992 and then rebase this one and test.

Joao-L-S-Almeida: Yes.

Joao-L-S-Almeida: This could be the solution: https://github.com/terrastackai/terratorch/pull/995

romeokienzler: @Joao-L-S-Almeida @christian-pinto 
@paolo-fraccaro and I will create a pypi package 4 GBv2 this week, so we can properly test everything, also #995 should be merged today or tomorrow, kudos to @Joao-L-S-Almeida

romeokienzler: @christian-pinto #995 has been merged

tox.ini
```diff
@@ -0,0 +1,13 @@
...
+    -e .[wxc,test]
```
christian-pinto:
```suggestion
    .[geobenchv2,wxc,test]
```

geobenchv2 is already a dependency in terratorch and no need for editable install.

christian-pinto (changes requested):

christian-pinto:
Apart for the above changes it all looks good to me.

tox.ini
Joao-L-S-Almeida: Yes, that's right.

Joao-L-S-Almeida: Done.

tox.ini
```diff
@@ -0,0 +1,12 @@
...
+requires =
+    tox >= 4.20
+env_list =
+    3.13
+    3.12
+    3.11
+
+[testenv]
+deps =
+    -e .[wxc,test,geobenchv2]
```
christian-pinto:
```suggestion
    .[wxc,test,geobenchv2]
```

I suggest removing the `-e` as you don't really need an editable install

Joao-L-S-Almeida: All right.

PR #995, opened 2025-11-07
Updating `segmentation-pytorch-models`

Joao-L-S-Almeida:
Geobench requires `torchgeo>=0.7.1` and that, in turn requires `smp>=0.5.0`. 

* Some parameters have changed its name, as `use_batchnorm`, now called `use_norm` and supporting strings and dictionaries as input. 
* The `forward` for the SMP decoders now receives lists (and not arguments) of tensors. 
* Even keeping the same choices for the parameters, the dimensions aren't exactly the same for some cases and the tests must be adapted. One breaking change seems to happens in `segmentation_models_pytorch/decoders/unet/decoder.py::UnetDecoder.forward`, but it isn't clear why it could change the configuration. 
* I'm skipping some models (maybe it's a matter of removing them in the future) considering the `smp==0.5` release notes (sic): *timm- prefixed encoders are deprecated, tu- variants are now the recommended way to use encoders from the timm library. Most of the timm- encoders are internally switched to their tu- equivalent with state_dict re-mapping (backward-compatible), but this support will be dropped in upcoming versions*.
* In some cases, `ruff` has modified the format to adapt it to PEP8.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ dependencies = [
   "jsonargparse",
   "torchgeo>=0.7.0,<0.7.2",
   "einops",
-  "timm>=0.0.15",
+  "timm>=1.0.15",
   "pycocotools",
   "huggingface_hub",
   "tifffile",


tests/test_decoders.py
Joao-L-S-Almeida, line 1: It was adjusted to follow the new behaviour of `smp.UnetDecoder`.

terratorch/models/decoders/unet_decoder.py
```diff
@@ -39,4 +46,4 @@ def __init__(
     def forward(self, x: list[torch.Tensor]) -> torch.Tensor:
         # The first layer is ignored in the original UnetDecoder, so we need to duplicate the first layer
         x = [x[0].clone(), *x]
-        return self.decoder(*x)
+        return self.decoder(x)
```
romeokienzler, line 49: @Joao-L-S-Almeida can you please explain why this change was necessary?

terratorch/registry/smp_registry.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, decoder, out_channels: int) -> None:
         self.out_channels = out_channels
 
     def forward(self, x):
-        return self._smp_decoder(*x)
+        return self._smp_decoder(x)
```
romeokienzler, line 40: @Joao-L-S-Almeida can you please explain why this change was necessary?

Joao-L-S-Almeida: Now the forward method of these models receives the entire list of embeddings, not the elements as arguments.

terratorch/models/decoders/unet_decoder.py
Joao-L-S-Almeida: Now the forward method of these models receives the entire list of embeddings, not the elements as arguments.

PR #996, opened 2025-11-07
Separating vllm dependencies from tests dependencies

christian-pinto:
This PR creates two new optional dependencies groups:
- `vllm`: it only installs vLLM 0.11.0. This should be only installed by people using terratorch + vLLM
- `vllm_test`: it installs the packages for testing the vLLM io_processor plugin in Terratorch

The benefit is that the main test pipeline is currently separated from the vLLM one. This means that so far we have been installing vLLM for all tests even though they were not required. Also, it allows people who want to use vLLM + TT to quickly only get the necessary libraries for running their applications.

### DIFF
--- a/integrationtests/vllm_tests_set.sh
+++ b/integrationtests/vllm_tests_set.sh
@@ -13,7 +13,7 @@ source ${VENV_NAME}/bin/activate
 
 cd terratorch
 
-uv pip install --no-cache-dir -e .[test]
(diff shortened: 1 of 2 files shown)


christian-pinto:
This PR depends on #989 and https://github.com/terrastackai/terratorch/issues/992 and it forces geobenchv2 to be installed together with vLLM.

PR #998, opened 2025-11-12
Fix Geobench2, compute-statistics, multimodal transform

blumenstiel:
Fix main...

Remove try catch from geobench2, fix multimodal transform, fix nan values in images for compute_statistics


Joao-L-S-Almeida:
Are you tests still failing for the latest commit: 38a598892cb0f9c8d18b9a6a5535218f3ce18084, @blumenstiel ?
I think the try-catch would be necessary for cases in which the user doesn't want geobench, since it's optional.

PR #1005, opened 2025-11-13
Mute albumentations

blumenstiel:
Needs to be done in init.py via a env var. overwriting the logger did not work


Joao-L-S-Almeida:
LGTM.

PR #1026, opened 2025-11-25
add tests embedding generation

Isabelle-Wittmann:
Add tests for the current scope of the embedding generation task


Isabelle-Wittmann:
Part of #PR1048

Issue #1019, opened 2025-11-24
CrossEntropyLoss incompatible with default `ignore_index` parameter.


Foxigod:
I believe the removal of the NoneType check of the `ignore_index` parameter:
https://github.com/terrastackai/terratorch/blob/a65d86a8470a040e2079d5524eca237d107a9ff0/terratorch/tasks/segmentation_tasks.py#L222-L224
in the following commit: https://github.com/terrastackai/terratorch/commit/5afdc22aedc7c19a464bd97cd1ef9607ede7da56 (cc. @blumenstiel) resulted in the `ignore_index` parameter being 'broken' (I mention this in the following comment on said commit).

> I believe the removal of the None-type check on the `ignore_index` (which replaced None with -100) can be a cause for issues for the CrossEntropyLoss, which seems to not handle NoneType `ignore_index`, and instead require an integer one if provided.
> 
> I'm encountering these issues:
> ```python
>   [...]
>   File "VENV/lib/python3.12/site-packages/terratorch/tasks/segmentation_tasks.py", line 389, in validation_step
>     loss = self.val_loss_handler.compute_loss(model_output, y, self.criterion, self.aux_loss)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "VENV/lib/python3.12/site-packages/terratorch/tasks/loss_handler.py", line 45, in compute_loss
>     loss = self._compute_loss(model_output.output, ground_truth, criterion)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "VENV/lib/python3.12/site-packages/terratorch/tasks/loss_handler.py", line 78, in _compute_loss
>     loss: Tensor = criterion(y_hat, ground_truth)
>                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "VENV/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl>
>     return self._call_impl(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "VENV/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1520, in _call_impl
>     return forward_call(*args, **kwargs)
[...]

PR #1027, opened 2025-11-26
Fixed ignore_index NoneType for CE loss

blumenstiel:
See issue #1019

PR #1028, opened 2025-11-26
Fixed loss selection in task

blumenstiel:
Based on #1019, I reviewed the loss selection on the TT tasks. In classification I removed the `ignore_index` as it does not make sense. Fixed some imports in the multilabel classification task.

### DIFF
--- a/terratorch/tasks/regression_tasks.py
+++ b/terratorch/tasks/regression_tasks.py
@@ -183,51 +183,40 @@ def get_module_and_class(path):
     return path_, class_name
 
 
-def init_loss(loss: str, ignore_index: int = None, custom_loss: bool = False, custom_loss_kwargs: dict = None):
-    if custom_loss:
-        assert custom_loss_kwargs, "If you are using a custom loss, the `custom_loss_kwargs` are required."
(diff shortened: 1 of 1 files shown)


blumenstiel:
This pr is based on #1027 which fixes an issue in the segmentation task only.

Foxigod:
Why do you say that the `ignore_index` does not make sense for classification tasks, is it because the user could/should have already removed the samples in the dataloader? If that is the case, wouldn't it make more sense to notify the user that they are conducting bad practice instead of removing previous functionality and only providing the user with a warning?

blumenstiel:
Yes, exactly. In classification you only have one target value per sample, so this should be an ignore value. I am not sure if the previous functionality even worked because probably nobody has tested it. If we allow ignore values, we need to test if the loss returns NaN or 0

blumenstiel:
I just tested: If I only input ignore values, the ce loss is NaN, for smp losses it's empty. If only some samples in a batch are ignore values, torch and smp ignore them during reduction. I think we can assume that at least one valid sample per batch can be expected, so it would work to keep it. I add it again

romeokienzler:
rebased and resolved conflicts, testing...

PR #1042, opened 2025-12-12
fixed issues with the tutorial. missing model and broken file import

Mdmdma:
Fixed issues with the tutorial. The selected model does not exist any more on hugging face, it was replaced with a more recent prithvi model. Additionally the templating to find the images was broken (lines 544ff), this is fixed and the parameters made explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Developers, check out the [architecture overview](https://terrastackai.github.io
 
 This project welcomes contributions and suggestions. Ways to contribute or get involved:
 
-- Join our [Discord](https://discord.gg/x8AWY8QQ)
+- Join our [Discord](https://discord.gg/vQXTNmrkTM)
 - Create an [Issue](https://github.com/terrastackai/terratorch/issues) (for bugs or feature requests)
 - Contribute via [PR](https://github.com/terrastackai/terratorch/pulls)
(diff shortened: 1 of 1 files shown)


examples/notebooks/Tutorial.ipynb
```diff
@@ -15,9 +15,8 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "/home/jalmeida/.pyenv/versions/3.11.10/envs/terratorch/lib/python3.11/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html\n",
-      "  from .autonotebook import tqdm as notebook_tqdm\n",
-      "INFO:albumentations.check_version:A new version of Albumentations is available: 2.0.3 (you have 1.4.10). Upgrade using: pip install --upgrade albumentations\n"
+      "/scratch3/merler/miniconda/envs/terra/lib/python3.12/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html\n",
```
romeokienzler: can we please remove unnecessary output throughout the whole notbook? will also make it more readable...

romeokienzler:
@Mdmdma thanks so much for the pr, added a comment

examples/notebooks/Tutorial.ipynb
Mdmdma: @romeokienzler thank you for the review. I removed all output from the notebook. I have to say though that in case the notebook is not functioning properly any more it can be nice to see what a normal output would look like.

romeokienzler (approved):

romeokienzler:
Thanks @Mdmdma 

- [ ] merged

You are right, you could see the output in case the notebook is defunct, but then there is no us of it anyway IMHO, so what we are planning to do for TerraTorch 1.3 is having all Tutorials included in the automatic testing. We know that in 1.2 we are breaking most of the examples so we are deprecating

PR #1047, opened 2025-12-17
release v1.2, deprecate examples

romeokienzler:


examples/alexnet_custom_model.py
```diff
@@ -1,54 +1,5 @@
-import os
-import sys
-import torch
-import terratorch
-import warnings
-import os
-import sys
+Examples are deprecated since they are not compatible with TerraTorch >= v1.2 anymore - we're urgently working on a new set of nice and clean examples for you
```
christian-pinto: Why do you want to keep these files if you are moving the examples in the deprecated folder? I would just remove these files and add a  line in the changelog and/or docs saying that examples got moved nito deprecated folder and they will be removed soona nd replaced with new ones.

romeokienzler: the problem is that we don't know who is linking to those files from tutorials and we don't wanna create dead links

romeokienzler: will change the python files to compile and print the message

christian-pinto: OK, that's better. I suggest also replacing all those text messages into comments both in .py and .yaml files if they are not going to get printed.

romeokienzler: @christian-pinto done

christian-pinto (approved):
LGTM

christian-pinto (approved):
LGTM

PR #1048, opened 2025-12-17
Embedding Workflows Integrations

Isabelle-Wittmann:
This PR updates the _embedding generation task_, implements additional functionalities needed for _embedding downstream task_ workflows and adds sample notebooks for both.

- **Add Identity Backbone**  
  Introduces a simple backbone that directly passes input features to the necks. This is used for embedding downstream task workflows where no backbone is required.

- **Global `image_size_out` for pixelwise models**  
  Adds a global `image_size_out` feature required for embedding downstream workflows for pixelwise tasks.  
  The value is resolved in the forward pass with the following priority:
  1. Global `self.image_size_out`
  2. `kwargs.get("image_size")`
  3. Inferred via `get_image_size(x)`

- **Extend plotting to embedding PCA**  
  Adds embedding plotting (eg. during training) used in pixelwise tasks.  
  Introduces two global features to generic pixelwise data module:
  - `embedding_input` (bool) toggles between the existing plotting behavior and embedding-based plotting (defaults to `False`)
  - `pca_step`: subsampling factor to speed up PCA by using only a subset of pixels (only used when `embedding_input=True`, with a safe default)  
  Embedding plotting uses a PCA-based RGB proxy image, rescales it to the spatial resolution of the GT masks, and slightly adapts plotting styles and titles.  
  To simplify the code structure, both the new `to_pca_rgb` logic and the existing `to_rgb` logic were moved as helpers to `terratorch/datasets/utils.py`.

- **Make means and stds optional**  
  The pixelwise datamodule previously required `means` and `stds`, which is annoying for high-dimensional embeddings that don’t need normalization.  
  These arguments are now optional; if not provided, inputs are not normalized.

- **Ignore `NotGeoreferencedWarning` in `__getitem__`**  
I had persisting issues with global ignoring of NotGeoreferencedWarnings, for me it only works to add it to the file loading method as the setting gets overwritten by rasterio otherwise

- **Update embedding generation**  
  Introduce a new `EmbeddingOutputModel` that wraps backbones with necks for embedding generation and is available as an alternative model type in `EncoderDecoderFactory`.  Therefore decoder input to`EncoderDecoderFactory` is now optional; behavior for other model types remains unchanged.

  Updates to necks needed to make this work:
  - **AggregateTokens**
    - Now supports applying aggregation independently to multiple layer outputs  
    - `index` is deprecated but remains backward compatible  
    - New `indices` argument supports a single int or a list of ints  
    - Adds `drop_cls` to control dropping the first token for `("mean", "min", "max")` aggregation  
    - Supports passing through 5D spatiotemporal inputs (e.g. with `TemporalWrapper`, `pooling="keep"`), preserving the temporal dimension
  - **ReshapeTokensToImage**
    - Passes inputs with more than 3 dimensions (4D spatial or 5D spatiotemporal) through unchanged  
    - The original 3D token reshaping logic remains unchanged

  Embedding generation task updates:
  - Uses `EmbeddingOutputModel` via `EncoderDecoderFactory`
  - Necks are appended at build time based on selected layer indices, pooling strategy, and output type  
    (for TIFF outputs, embeddings are reshaped to spatial outputs). This changes the expected inputs a bit.
  - Automatically writes a config file alongside the embedding output directory to document chosen specifications and folder name explanations (How layer index corresponds to actual model layer)
  - Adds a new `parquet_joint` output format that produces a single parquet file for all samples (only works for aggregated 1D embeddings)

- **Add embedding generation tests**  
  Updates and extends the embedding generation tests (building on PR #1026) to reflect the updates above.

- **Add embedding workflow tutorials**  
  Adds two new example notebooks (embedding generation and embedding downstream tasks, both on BurnScars data) and updates existing sample config to reflect the new changes.


terratorch/datamodules/generic_pixel_wise_data_module.py
```diff
@@ -173,6 +175,8 @@ def __init__(
                 Defaults to None, which simply applies ToTensorV2().
             no_data_replace (float | None): Replace nan values in input images with this value. If none, does no replacement. Defaults to None.
             no_label_replace (int | None): Replace nan values in label with this value. If none, does no replacement. Defaults to None.
+            embedding_input (bool): Whether the input represents embeddings rather than an image, used for plotting. Defaults to False.
+            pca_step (int): Spatial downsampling factor used when fitting PCA for embedding viusalisations. Defaults to 4 -> 1/4 of spatial embeddings used.
```
blumenstiel: Can you add some more explanation? I don't really understand what pca_step is doing. Is it about spatial upsampling?

terratorch/datasets/generic_pixel_wise_dataset.py
```diff
@@ -80,6 +84,8 @@ def __init__(
                 Defaults to None, which simply applies ToTensorV2().
             no_data_replace (float | None): Replace nan values in input images with this value. If none, does no replacement. Defaults to None.
             no_label_replace (int | None): Replace nan values in label with this value. If none, does no replacement. Defaults to -1.
+            embedding_input (bool): Whether the input represents embeddings rather than an image, used for plotting. Defaults to False.
+            pca_step (int): Spatial downsampling factor used when fitting PCA for embedding viusalizations. Defaults to 4 -> 1/4 of spatial embeddings used.
```
blumenstiel: same

terratorch/datamodules/generic_pixel_wise_data_module.py
Isabelle-Wittmann: Will do, its the portion of "pixels/spatial embeddings" used for PCA, 4 uses only every fourth element and is therefore more efficient, quite common to do and I tested it quite a bit, looks nearly identical to PCA on all pixels even up to 32

terratorch/datasets/generic_pixel_wise_dataset.py
```diff
@@ -182,11 +186,86 @@ def __getitem__(self, index: int) -> dict[str, Any]:
         return output
 
     def _load_file(self, path, nan_replace: int | float | None = None) -> xr.DataArray:
+        warnings.filterwarnings("ignore", category=NotGeoreferencedWarning)
```
blumenstiel: I know these are super annoying but executing this in every getitem step is a bit redundant. Maybe something that we can do more high level, e.g. in cli_tools.py? Can you add a TODO that we have to solve that in another way?

terratorch/datasets/generic_pixel_wise_dataset.py
```diff
@@ -182,11 +186,86 @@ def __getitem__(self, index: int) -> dict[str, Any]:
...
+        if isinstance(image, Tensor):
+            image = image.detach().cpu().numpy()
+
+        label_mask = sample["mask"]
+        if isinstance(label_mask, Tensor):
+            label_mask = label_mask.detach().cpu().numpy()
+
+        if embedding_input:
+            if image.ndim == 2:
+                return  # Plotting requires (reshaped) spatial (C, H, W) input embeddings.
```
blumenstiel, line 244: Maybe a `warnings.warn` that a neck is missing?

terratorch/datamodules/generic_pixel_wise_data_module.py
Isabelle-Wittmann: (used only for the computation of the PCA directions, then its applied to all pixels of course)

terratorch/datasets/generic_pixel_wise_dataset.py
Isabelle-Wittmann: Nothing worked for me, because I think rasterio in a similar way activates it again haha, we can also leave it out, but whats the downside of having it in every pass?

terratorch/datasets/utils.py
```diff
@@ -225,3 +227,62 @@ def clip_image_percentile(img: np.ndarray, q_lower: float = 1, q_upper: float =
...
+
+
+def to_rgb(image_chw: np.ndarray, rgb_indices: list[int]) -> np.ndarray:
+    """Convert a channel-first image (C, H, W) to an RGB image for visualization."""
+
+    img = image_chw.take(rgb_indices, axis=0)
+    img = np.transpose(img, (1, 2, 0))
+
+    ch_min = img.min(axis=(0, 1), keepdims=True)
+    ch_max = np.maximum(img.max(axis=(0, 1), keepdims=True), 1e-8)
```
blumenstiel: Should we do a scaling with np.quantiles([0.01, 0.99])? In my experience it works a bit better

Isabelle-Wittmann: Yes I can add it, this was just closer to the existing code

terratorch/datasets/utils.py
```diff
@@ -225,3 +227,62 @@ def clip_image_percentile(img: np.ndarray, q_lower: float = 1, q_upper: float =
...
+    emb_flat = image_chw.reshape(C, -1).T  # (N, C)
+
+    emb_fit = emb_flat[::step].astype(np.float32, copy=True)
+    mean = emb_fit.mean(axis=0, keepdims=True)
+    emb_fit -= mean
+
+    pca = PCA(n_components=3, svd_solver="randomized", random_state=0)
+    pca.fit(emb_fit)
+
+    proj = pca.transform(emb_flat.astype(np.float32, copy=False) - mean)  # (N, 3)
```
blumenstiel: Isn't there usually a `fit_transform` method in sklearn? Don't know but it might be a bit optimized

terratorch/models/embedding_output_model.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if neck is not None:
+            self.neck = neck
+
+        else:
+            self.neck = lambda x, image_size: x
+
+        self.patch_size = patch_size
+        self.padding = padding
+        self.encoder.eval()
```
blumenstiel: Might be better to run `freeze_encoder` by default and add a warning if it is false? 
Because I don't know if lightning somewhere runs model.train().

terratorch/models/embedding_output_model.py
```diff
@@ -0,0 +1,65 @@
...
+
+        self.encoder = encoder
+
+        if neck is not None:
+            self.neck = neck
+
+        else:
+            self.neck = lambda x, image_size: x
+
+        self.patch_size = patch_size
```
blumenstiel, line 45: Maybe get the `patch_size` from the models? See pixel wise model. Just to make it more user friendly

Isabelle-Wittmann: Im doing it the same way as pixelwise model: Patch size is fetched from backbone in EncoderDecoderFactory and then passes it on as argument to either pixelwise model or embedding output model - or do you mean something else

terratorch/models/embedding_output_model.py
Isabelle-Wittmann: Adapted

terratorch/datasets/utils.py
Isabelle-Wittmann: fit_transform only is simpler if all the pixels you want to transform are part of the set you use to fit PCA, for pca_step > 1 thats not the case

terratorch/datasets/utils.py
Isabelle-Wittmann: Adapted

terratorch/models/necks.py
```diff
@@ -65,40 +65,69 @@ def process_channel_list(self, channel_list: list[int]) -> list[int]:
...
+                T = feat.shape[2]
+                feat = feat.reshape(B, -1, T, self.latent_dim[i])
+
+            if isinstance(self.pooling, int):
+                # Select token index
+                aggregated_features.append(feat[..., self.pooling, :])
+            elif self.pooling == "cls":
+                # Assuming CLS token is on first position
+                aggregated_features.append(feat[..., 0, :])
+            elif self.drop_cls:
```
blumenstiel: Maybe run this before all the elif pooling parts and ensure in the init that drop_cls and pooling == cls is not possible?

This way they code looks a bit nicer:)

terratorch/tasks/embedding_generation.py
```diff
@@ -38,38 +44,65 @@ def __init__(
...
-            raise ValueError(f"Unsupported output format: {self.output_format}. Supported formats are 'tiff' and 'parquet'.")
-        
-        if self.embedding_pooling is not None:
-            if self.has_cls is None and self.embedding_pooling.startswith("vit_"):
-                warnings.warn("No 'has_cls' provided; assuming CLS if token count is odd.")
+        self.embedding_indices = layers
+
+        model_args = model_args or {}
+        if embedding_pooling in (None, "None", "keep"):
+            model_args["necks"] = [
```
blumenstiel: Maybe check if necks are not defined by the user before writing it?

terratorch/datasets/generic_pixel_wise_dataset.py
Isabelle-Wittmann: Adapted - however I had to add a warning that spatial arranged input embeddings are needed (add a Neck during embedding generation). With our current plotting functionality we can only use the input not an intermediate representation -> good future ToDo

terratorch/tasks/embedding_generation.py
Isabelle-Wittmann: Adapted

terratorch/models/necks.py
Isabelle-Wittmann: Adapted

terratorch/models/embedding_output_model.py
blumenstiel: Ah sorry, it's handled in EncDecFactory, so it also works for the generation.

terratorch/datasets/utils.py
blumenstiel: Ah I see

terratorch/datasets/generic_pixel_wise_dataset.py
blumenstiel: Maybe it's because we load the data via rioxarray? Using rasterio directly might work.

terratorch/datamodules/generic_pixel_wise_data_module.py
blumenstiel: That makes sense, smart idea:)

terratorch/datasets/generic_pixel_wise_dataset.py
Isabelle-Wittmann: No, doesn't work as well. I comment the above warning out for now, and take another look in the next few days

romeokienzler:
@blumenstiel merging ok from ur pov?

blumenstiel (approved):

PR #1066, opened 2026-01-05
add augmentation for object detection plus example

romeokienzler:


Isabelle-Wittmann:
@romeokienzler please stick to the new example structure and put this under `examples/object_detection`

romeokienzler:
@Isabelle-Wittmann sorry for that, fixed in #1100

PR #1069, opened 2026-01-08
Fix TM caption tokenizer

blumenstiel:


romeokienzler:
@blumenstiel tests operational again, on it

PR #1082, opened 2026-01-22
Revert "Fixed loss selection in task"

romeokienzler:
Reverts terrastackai/terratorch#1028


romeokienzler:
@blumenstiel can you please approve the revert PR as I've accidentally merged it wo/ running the full test suite but only a single test which failed before...

PR #1083, opened 2026-01-23
Fix multitemporal dataset

Isabelle-Wittmann:
Fix MultiTemporalCropClassification Dataset rearrange bug (https://github.com/terrastackai/terratorch/issues/1080) and update generic datamodules

### DIFF
--- a/terratorch/datasets/generic_multimodal_dataset.py
+++ b/terratorch/datasets/generic_multimodal_dataset.py
@@ -174,9 +174,9 @@ class GenericMultimodalDataset(NonGeoDataset, ABC):
                     "(all timesteps of one band are stacked together), "
                     "set temporal_channel_major=True.")
 
-            if output_bands is None:
+            if dataset_bands is None:
                 raise ValueError(
-                    "Please provide output_bands when expand_temporal_dimension=True."
+                    "Please provide dataset_bands when expand_temporal_dimension=True."
                 )
 
         if scalar_label:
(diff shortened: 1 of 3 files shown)


Isabelle-Wittmann:
Also updated to address https://github.com/terrastackai/terratorch/issues/1003 -> the rearrange needs the channel number from the input bands not output bands given its applied before filtering

romeokienzler:
testing on ccc...

terratorch/datasets/generic_multimodal_dataset.py
```diff
@@ -162,9 +165,19 @@ def __init__(
...
+        if self.expand_temporal_dimension:
+            if not self.temporal_channel_major:
+                warnings.warn(
+                    "expand_temporal_dimension=True assumes bands are grouped by time "
+                    "(all bands of one timestep are stacked together). "
+                    "If instead bands are grouped by channel "
+                    "(all timesteps of one band are stacked together), "
+                    "set temporal_channel_major=True.")
+
+            if output_bands is None:
```
blumenstiel: Does it have to be dataset bands or outputs bands? Because the check was with dataset bands before?

blumenstiel: Code below uses dataset bands

blumenstiel:
@romeokienzler how did the tests go?

terratorch/datasets/generic_multimodal_dataset.py
Isabelle-Wittmann: Yes you are right! Thanks a lot for catching that ...  I first updated the other two generic datamodules which (wrongly) used output bands, before I changed that

PR #1095, opened 2026-02-04
Remove scale_modules from UPerNet decoder

romeokienzler:
Removed deprecated scale_modules parameter and related functionality from the UPerNet decoder.


romeokienzler:

Ran on our IBM BlueVela LSF cluster (fresh venv, `pip install -e ".[test]"`, Python 3.12).

**Code assessment.** Removes the deprecated `scale_modules` path (and the `fpn1..4` blocks) from `UperNetDecoder`, pointing users to the `LearnedInterpolateToPyramidal` neck instead. Direction is good, but there are three things to resolve before merge:

1. **Merge conflict with `main`.** The branch currently conflicts in `terratorch/models/decoders/upernet_decoder.py` (GitHub shows `CONFLICTING`, and a local `git merge origin/main` fails there). Needs a rebase/merge.

2. **Callers still pass the removed kwarg → `TypeError`.** `extract_prefix_keys(kwargs, "decoder_")` forwards `decoder_scale_modules`/`scale_modules` straight into `UperNetDecoder.__init__`, so removing the parameter breaks existing callers. Concretely, these will fail at model-build time (all guarded on `decoder == "UperNetDecoder"`):
   - `tests/test_prithvi_model_factory.py` (sets `decoder_scale_modules = True`, 4 sites)
   - `tests/test_clay_tasks.py` (sets `scale_modules = True`, 3 sites)
   - `examples/datasets_and_benchmarks/chesapeake_dataset_prithvi.yaml:90` (`decoder_scale_modules: true`)
   These need to be updated in the same PR (migrate them to `LearnedInterpolateToPyramidal`, or drop the kwarg).

3. **Dead import.** After removing the `warnings.warn(...)` deprecation block, `import warnings` at the top of `upernet_decoder.py` becomes unused (F401).

**Unit tests** — `tests/test_decoders.py`: **8 passed** ✅ (note: this file does *not* cover the factory/kwarg path in item 2, which is where the regression lives.)

**Integration (base set):** could not run against a `main` merge because of the conflict in item 1. On the raw branch, `test_models_fit` fails only due to the pre-#1206 hardcoded `/dccstor` data paths (environment artifact, not this change). Once the conflict and the caller updates above are resolved, I'll re-run the merged base set.

rosielickorish (approved):
LGTM

PR #1102, opened 2026-02-11
Embedding Georeferencing Updates

Isabelle-Wittmann:
Updates to handle georeferencing during embedding generation better, this mostly includes

1) Updates to `EmbeddingGenerationTask`
- Now uses georeferencing metadata in various, prioritized ways for geotiff and geoparquet outputs, eg. for geotiff prefer geotransform+ raster shape, otherwise try bounds, or lastly infer from center coordinates. If available metadata can not be used for georeferencing it's still saved as tags/ columns. 
- Also includes updates for joint_parquets, that are now saved as part files and joint at the end of predict epoch -> this also cause a minor update to the embedding generation tests

2) Updates that were needed outside of `EmbeddingGenerationTask`
- Update pixelwise datamodule to be able to return georeferencing metadata in forward pass for georeferenced tiff inputs, handled why new init flag `return_georeference`, which defaults to `False` so nothing changes for existing workflows. 

3) Other fixes covered in this PR:
- We changed to using existing TT necks for embedding generations (#1048 ), which causes a problem for temporal input data that is passed through (w/o temporal pooling), `ReshapeTokensToImage` now also works for inputs with a separate time dim
- Also noticed a bug in the edge case for `ReshapeTokensToImage` 
- Minor update to an overlooked docstring for embedding model

### DIFF
--- a/integrationtests/vLLM/test_segmentation_io_processor.py
+++ b/integrationtests/vLLM/test_segmentation_io_processor.py
@@ -77,7 +77,7 @@ def test_serving_segmentation_plugin(get_server, model_name, input_name):
         # This is just in case the test ends up with a GPU of less memory than an A100-80GB.
         # Just to avoid OOMing in the CI
         "--max-num-seqs",
-        "8",
+        "32",
         "--io-processor-plugin",
         io_processor_plugin,
         "--model-impl",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ include = ["terratorch*"]
 
 [project]
 name = "terratorch"
-version = "v1.2.1"
+version = "v1.2.2"
 description = "TerraTorch - The geospatial foundation model fine-tuning toolkit"
 license = { "text" = "Apache License, Version 2.0" }
 readme = "README.md"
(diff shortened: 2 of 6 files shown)


tests/test_embedding_generation.py
```diff
@@ -366,19 +366,13 @@ def test_write_parquet_batch_and_close_parquet(temp_dir, mock_backbone_registry)
...
-        dir_path=Path(temp_dir),
+        dir_path=Path(temp_dir) / "layer_00",
     )
+    task.join_parquet_files()
 
-    out = Path(temp_dir) / "embeddings.parquet"
+    out = Path(temp_dir) / "layer_00" / "embeddings.parquet"
     assert out.exists()
 
-    # close resets writer + path
```
romeokienzler, line 375: @Isabelle-Wittmann why have those been removed?

Isabelle-Wittmann: Because the parquet logic in embedding_generation changed, before we used a parquet_writer and path that remained open and appended new rows to the parquet file, however this is not so efficient, we've now switched to write parquet_parts files and having a final join_parquet files call that joins them into one, and deletes the parts files. Long story short: embedding generation task doesnt have the features anymore, making the assert non-necessary. The join_parquet logic is tested by the out.exists()

PR #1104, opened 2026-02-11
Update docs from main

blumenstiel:


romeokienzler:
@blumenstiel thanks a lot, we've decided to keep docs in sync with main and when we release we just tag, will delete docs branch once we see everything runs stable, updated the wf in #1117 - therefore closing

PR #1108, opened 2026-02-16
add custom modules example

romeokienzler:


examples/custom_modules/config.yaml
```diff
@@ -0,0 +1,35 @@
+trainer:
+  accelerator: auto
+  devices: 1
+  max_epochs: 2
+  log_every_n_steps: 5
+
+custom_modules_path: ./custom_modules
```
christian-pinto, line 7: Can I set this to any folder?

In the example notebook the config is never loaded, so it is not clear when is terratorch loading the modules located at `custom_modules_path`

examples/custom_modules/config.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+model:
+  class_path: terratorch.tasks.regression_tasks.PixelwiseRegressionTask
+  init_args:
+    model_factory: EncoderDecoderFactory
+    model_args:
+      backbone: HelloGeoModule
```
christian-pinto, line 14: Not clear how the classname is resolved by TT at runtime.

christian-pinto (changes requested):
I suggest adding a README that explains how to setup the whole thing, where to place the files, etc.

examples/custom_modules/config.yaml
romeokienzler: custom_modules is the default, so if you don't set it, it will be used

examples/custom_modules/config.yaml
romeokienzler: Added the explaination to the [docs](https://github.com/terrastackai/terratorch/pull/1108/changes#diff-c2a1915ea99f7e99a9573dcdb452852244a77cc94e132acd6457198dc3dacdfdR111)

romeokienzler:
@christian-pinto 
> I suggest adding a README that explains how to setup the whole thing, where to place the files, etc.

done

examples/custom_modules/config.yaml
christian-pinto: so by default, when starting the registry it will search for custom modulesin that location?

I see now, from here

https://github.com/terrastackai/terratorch/blob/d26539b692f314f580296d123189b6876715e3d7/terratorch/registry/custom_registry.py#L6-L15

it loads from `$PWD/custom_modules` by default

.github/workflows/auto-approve-docs.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: tj-actions/changed-files@v42
+        with:
+          files_ignore: |
+            **/*.md
+
+      - name: Approve PR
+        if: steps.changed-files.outputs.any_changed == 'false'
+        uses: hmarr/auto-approve-action@v4
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
christian-pinto, line 28: This file seems not to belong to this PR

examples/custom_modules/README.md
```diff
@@ -0,0 +1,63 @@
...
+examples/custom_modules/
+├── config.yaml
+└── custom_modules/
+    ├── __init__.py
+    └── hello_geo_module.py
+```
+
+## Why `__init__.py` matters
+
+TerraTorch imports the package from `custom_modules_path`. Registration decorators run only when the module is imported.
```
christian-pinto: just for the sake of helping the reader, I would explain the TT loads from `$PWD/custom_modules` by default.

christian-pinto (changes requested):

.github/workflows/auto-approve-docs.yml
romeokienzler: yes, I've piggybacked this in to allow for documentations updates without approval

examples/custom_modules/README.md
romeokienzler: updated

christian-pinto (approved):
LGTM

.github/workflows/auto-approve-docs.yml
adamjstewart, line 1: I would strongly recommend reverting this. I understand the desire to save yourself one more click, but you are also saving attackers one more click. This can lead to a more serious supply chain vulnerability if you aren't careful...

PR #1152, opened 2026-03-22
Fix chained generations in TerraMind

blumenstiel:
Due to missing encoder embeddings, TerraMind did not use chained generations in TerraTorch.


terratorch/models/backbones/terramind/model/terramind.py
```diff
@@ -857,6 +858,42 @@ def build_modality_embeddings(modality_info, modalities, img_size=None, dim=None
...
+                                )
+
+            # Get modality key in MODALITY_INFO
+            if modality in modality_info.keys():
+                key = modality
+            elif 'sen2' in modality_renamed:
+                key = 'tok_sen2l2a@224'
+            elif 'sen1rtc' in modality_renamed:
+                key = 'tok_sen1rtc@224'
+            elif 'sen1' in modality_renamed:  # Default to S1GRD if not specified
```
Isabelle-Wittmann: I would update the sen1 naming, by a) adding `sen1grd` or `sen1` here and b) updating the block above to also work for `sen1`

terratorch/models/backbones/terramind/model/terramind.py
Isabelle-Wittmann: similar to comment below, I would add `sen1` as default

Isabelle-Wittmann:
Looks good, pls check the naming conventions as commented

Isabelle-Wittmann (dismissed):

terratorch/models/backbones/terramind/model/terramind.py
blumenstiel: Good point, thanks! I changed it to `sen1`.

terratorch/models/backbones/terramind/model/terramind.py
blumenstiel: `sen1` basically covers `grd` as well:)

Isabelle-Wittmann (dismissed):

terratorch/models/backbones/terramind/model/terramind.py
```diff
@@ -857,6 +858,42 @@ def build_modality_embeddings(modality_info, modalities, img_size=None, dim=None
...
         mod_name_mapping[modality] = key  # Requires manual mapping for loading model weights
 
+    # Conditioned modalities
+    if cond_modalities:
+        for modality in cond_modalities:
+            # Cover multiple naming conventions
+            modality_renamed = (modality.lower()
+                                .replace('s2', 'sen2')
+                                .replace('s1', 'sen1')
+                                .replace('text', 'caption')
```
jhnnsjkbk: is that needed?

blumenstiel: It gives the user more flexibility for naming the modalities. E.g. `S2L2A` would otherwise fail.

jhnnsjkbk (approved):
All good. Let's think if we want to remove captions to avoid misunderstandings?

terratorch/models/backbones/terramind/model/terramind.py
jhnnsjkbk: Sorry, just meant the text & caption part

blumenstiel:
Agree, but then v0.1 would not work anymore. But I can at least remove it here so people need to call it caption

jhnnsjkbk1 (approved):

Isabelle-Wittmann:
@blumenstiel ready to merge?

PR #1160, opened 2026-03-26
Add RandomForest decoder and embedding classification datamodule

arishofmann:
Here is my PR for the RF decoder and embedding classification datamodule I added

### DIFF
--- a/terratorch/datamodules/__init__.py
+++ b/terratorch/datamodules/__init__.py
@@ -8,6 +8,9 @@ from terratorch.datamodules.generic_pixel_wise_data_module import (
     GenericNonGeoPixelwiseRegressionDataModule,
     GenericNonGeoSegmentationDataModule,
 )
+from terratorch.datamodules.embedding_classification_data_module import (
+    EmbeddingClassificationDataModule,
+)
 from terratorch.datamodules.generic_scalar_label_data_module import (
     GenericNonGeoClassificationDataModule,
 )
@@ -77,6 +80,7 @@ except ImportError:
     pass
 
 __all__ = [
+    "EmbeddingClassificationDataModule",
     "GenericNonGeoSegmentationDataModule",
     "GenericNonGeoPixelwiseRegressionDataModule",
     "GenericNonGeoClassificationDataModule",
--- a/terratorch/datamodules/embedding_classification_data_module.py
+++ b/terratorch/datamodules/embedding_classification_data_module.py
@@ -1,0 +1,68 @@
+# Copyright contributors to the Terratorch project
+
+import lightning
+from torch import Tensor
+from torch.utils.data import DataLoader, Dataset
+
+from terratorch.datasets.embedding_dataset import EmbeddingDataset
+
+
+class EmbeddingClassificationDataModule(lightning.LightningDataModule):
+    """DataModule for pre-extracted embeddings stored as .pt files."""
+
+    def __init__(
+        self,
+        batch_size: int,
+        num_workers: int = 0,
+        train_data_root: str | None = None,
+        val_data_root: str | None = None,
+        test_data_root: str | None = None,
+        drop_last: bool = True,
+    ) -> None:
+        super().__init__()
+        self.batch_size = batch_size
+        self.num_workers = num_workers
+        self.train_data_root = train_data_root
+        self.val_data_root = val_data_root
+        self.test_data_root = test_data_root
+        self.drop_last = drop_last
+
+        self.train_dataset: Dataset | None = None
+        self.val_dataset: Dataset | None = None
+        self.test_dataset: Dataset | None = None
+
+    def setup(self, stage: str) -> None:
+        if stage in ["fit"] and self.train_data_root:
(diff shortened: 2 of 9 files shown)


Isabelle-Wittmann:
Hi @arishofmann, thanks a lot for opening a PR, next time feel free to also tag us as reviewers then we get a notification :) I have some high level questions:

- Why do we need an extra embedding dataset and loader? I would prefer to use the generic datasets for this instead, as far as I can see we only need to extend the generic dataset to also be able to read in the pt files in the expected format.
-  Looking at the current task implementation, I'm not sure the inheritance from the other TerraTorch Tasks/ BaseTask makes sense to me, I think it would be cleaner to keep the existing TerraTorchTask for all gradient-based training, and then have an alternative base task for other types of EmbeddingDecoding mechanisms like RF. For this it would be best to define a general EmbeddingDecodingTask (that does not inherit from TerraTorchTask), from which a specific classification task can inherit if needed. I don't see a need why it needs to be RandomForest specific, would be much better to eg. keep generic for other supervised sklearn methods.
- With the above changes also the actual decoder wrapper can become a bit simpler

What do you think?

arishofmann:
Hi @Isabelle-Wittmann, thanks for the review.

On the embedding dataset: yeah, the generic datasets assume image I/O (GeoTIFFs, bands, transforms, spatial dims), so I made a separate one for .pt tensors. But I agree, extending the generic dataset to also handle .pt inputs would be cleaner. I can work on that. 
On the task: I inherited from MultiLabelClassificationTask mainly to reuse the metrics and val/test logging. The trade-off was workarounds like automatic_optimization = False. A separate EmbeddingDecodingTask makes sense. Should it inherit from TerraTorchTask (reuses model factory and freeze logic) or directly from LightningModule? Just want to align before I refactor. 
On making it generic: good point, I will generalize the decoder wrapper to accept any sklearn estimator instead of hardcoding RF.

Isabelle-Wittmann:
Hi @arishofmann, great - thanks a lot! For the task module, for me it would make more sense to not inherit from a TerraTorchTask but build a separate new (base) task that fits the sklearn decoders better.

arishofmann:
Hi @Isabelle-Wittmann , I worked through your three points: 
- I made the decoder generic.  And renamed RandomForestDecoder to SklearnDecoder, it now takes any sklearn estimator as a string parameter. 
- I created a separate EmbeddingDecodingTask base class with an EmbeddingClassificationTask subclass. Both inherit directly from BaseTask instead of TerraTorchTask, so no gradient-based training assumptions. 
- For the dataset I kept EmbeddingDataset as a separate class. The workflow loads bulk .pt tensors rather than individual image files, so extending the generic datasets felt forced. Happy to revisit if you see a clean way to integrate it. 

Tested it on CCC with RF on my Finland embeddings and it runs through and produces validation metrics.

Isabelle-Wittmann:
Thanks a lot, looks good! I'll test it tomorrow and then merge :)

romeokienzler:
@Isabelle-Wittmann cn u pse test and merge?

romeokienzler:

Thanks @arishofmann! Ran on our IBM BlueVela LSF cluster (fresh venv, `pip install -e ".[test]"`, Python 3.12).

**Code assessment.** Adds a `SklearnDecoder` (RandomForest) plus `EmbeddingDataset` / embedding-classification datamodule and tasks for sklearn-based classification on frozen embeddings. Nice feature. Two things to flag:

1. **Security — please switch to `weights_only=True`.** In `terratorch/datasets/embedding_dataset.py`:
   ```python
   self.patch_ids = torch.load(self.data_root / "patch_ids.pt", weights_only=False)
   ```
   This reintroduces `weights_only=False`, which #1208 just removed repo-wide as a security hardening (arbitrary code execution on untrusted checkpoints). If `patch_ids.pt` only holds tensors/plain data, please use `weights_only=True` (the default). If it genuinely needs to deserialize non-tensor objects, could you add a short comment justifying it?

2. **No dedicated tests for the new modules.** The new decoder/dataset/tasks don't ship unit tests. Import + wiring checks pass (below), but I'd suggest a small test that fits the RandomForest decoder on a tiny synthetic embedding set so this path is covered in CI.

**Unit tests** — `tests/test_decoders.py`: **8 passed** ✅
**Import/wiring smoke** — `EmbeddingDataset`, `SklearnDecoder`, `EmbeddingClassificationTask`, `EmbeddingDecodingTask` all import cleanly ✅

**Integration (base set):** not applicable — this feature isn't part of `test_models_fit`.

**Note (unrelated):** 1 pre-existing collection error in `tests/test_geobench_v2_data_module.py` (fixed separately on `main`), not caused by this branch.

romeokienzler (approved):
Approving after maintainer review, the `weights_only=True` security fix, and new unit tests (`tests/test_embedding_classification.py`, 18 passing on the IBM BlueVela cluster + green in the fast `build` matrix). The failing `build (3.13)` in the tuning-toolkit workflow is a pre-existing failure on `main` (tortilla tests need `terratorch[tortilla]`/tacoreader, which that CI job doesn't install) — unrelated to this PR.

PR #1170, opened 2026-04-17
Enable compatibility with vLLM thread pool execution for IOProcessors

mgazz:
Works in combination with https://github.com/terrastackai/terratorch/pull/1166 to restore support to Terratorch IOProcessors in vLLM

### DIFF
--- a/terratorch/vllm/plugins/segmentation/segmentation_io_processor.py
+++ b/terratorch/vllm/plugins/segmentation/segmentation_io_processor.py
@@ -356,11 +356,6 @@ class SegmentationIOProcessor(IOProcessor):
         request_id: str | None = None,
         **kwargs,
     ) -> PromptType | Sequence[PromptType]:
-        # TESTED: This now works with the thread pool implementation.
-        # Previously, calling asyncio.run() would fail with "cannot run two event loops" error
-        # because vLLM's serving infrastructure runs in an event loop.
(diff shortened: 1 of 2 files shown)


christian-pinto:
@mgazz is this a breaking change for previous versions of vLLM? Also, the change is a bit counterintuitive as we are starting an asynchronous function from a synchronous one.

mgazz:
> @mgazz is this a breaking change for previous versions of vLLM?

I need to investigate this.

> Also, the change is a bit counterintuitive as we are starting an asynchronous function from a synchronous one.

Why it works? The reason why this work is that with the introduction of the Thread Pool in vLLM the pre processing runs in a separate process that is independent to the vLLM's engine loop. As a matter of fact it does not matter if we run the pre/processing using async or not. The code runs in isolation inside the process.

Why such implementation? moving from an asynchronous implementation to a synchronous one would require many changes to the current implementation. With this approach we change only the way we call the pre-processing but we keep the actual logic unchanged.

mgazz:
> @mgazz is this a breaking change for previous versions of vLLM?

Verified that these changes do not break `vllm==0.19.1`

christian-pinto (approved):

christian-pinto:
@romeokienzler you can merge it if it looks good for you too!

PR #1173, opened 2026-04-21
Fix denormalize

blumenstiel:
param denormalize was moved from call to init, moving it back.


blumenstiel:
@ahmedemam576 I saw that you made the change when moving the Normalize function from generic_pixel_wise_data_module.py to utils.py. 
Did you made other changes that used denormalize as a parameter in the init? I made both ways work now

romeokienzler (approved):

PR #1182, opened 2026-05-05
update docs deploy workflow

Isabelle-Wittmann:
Updates docs deploy workflow:
- Deploys docs to gh-pages when files in `docs/`,` mkdocs.yml` or docs workflow change on main under latest
- Deploys on releases under stable as before

Also updates both workflows to one gh-pages push instead of of using mike --push twice


rosielickorish (approved):
LGTM

PR #1186, opened 2026-05-08
Support CPU training, add installation commands, download data and embeddings if not present

romeokienzler:
Support CPU training, add installation commands, download data and embeddings if not present


Isabelle-Wittmann (dismissed):
Thanks! 

Please update the following:
- Please remove the allowed 1608 length in the data checks as this does not fit the data structure that has 804 samples
- Can you updated the workflow to the following please such that users don’t download all the data before noticing it was maybe just a path issue: 1. Check if data is available in the right format and length (the “old cell”), 2. if not available print a warning “Required data not found, please check paths and otherwise either run the embedding generation notebook, or alternatively download percomputed embeddings by running the download cell below”, 3. After that insert the data download cell
- For the final commit please keep the cell outputs with visualizations so people that just scroll over the notebook see the outputs

ahmedemam576:
I still see 2 issues that Isabelle mentioned:

1-In examples/embeddings/downstream_segmentation_burnscars.ipynb, the notebook downloads data before checking if paths/data are correct. It should check first, then show the download step only if data is missing.

2-The notebook outputs were cleared in examples/embeddings/downstream_segmentation_burnscars.ipynb.
Please keep the visualization outputs in the final commit so readers can see results without running everything.

ahmedemam576:
comments were addressed

ahmedemam576 (approved):
the comments made by Isabella are addressed

PR #1200, opened 2026-06-26
Fix: Resolve KeyError and NameError in TorchGeo backbones weight loading

clementgilli:
## Description
This PR fixes two bugs that prevent the initialization of several TorchGeo-based backbones when using `pretrained=True`.

### 1. Missing TorchGeo Bands (`KeyError`)
**Files modified:** `terratorch/models/backbones/torchgeo_resnet.py`, `terratorch/models/backbones/torchgeo_vit.py` and `terratorch/models/backbones/torchgeo_swin_satlas.py`
- **Issue:** When loading weights, TorchGeo models often define their Sentinel-2 bands using formats like `"B4"`, `"B3"`, or `"B8a"`. These keys were missing from the `look_up_table`, causing a `KeyError` to be raised during `get_pretrained_bands()`. 
- **Fix:** Added the legacy TorchGeo band formats (`"B1"` to `"B9"`, specifically including the lowercase `"B8a"` which was causing silent failures) to the `look_up_table`. I proactively applied this fix to the Swin Satlas backbone as well to prevent identical crashes.

### 2. Undefined Function (`NameError`)
**File modified:** `terratorch/models/backbones/torchgeo_vit.py`
- **Issue:** In the `load_vit_weights` function, the `elif weights is not None` block calls a `remove_keys()` function. This function is not defined anywhere in the file or imported, resulting in a `NameError: name 'remove_keys' is not defined`.
- **Fix:** Replaced the undefined `remove_keys` call with the explicit dictionary key deletion logic (`for k in ["head.weight", "head.bias"]: del ...`) that is already correctly implemented a few lines above in the same file.

## Testing
Tested locally. Models such as `seco_resnet18_sentinel2_rgb_seco` and `ssl4eos12_vit_small_patch16_224_sentinel2_all_dino` now successfully translate band names, process the checkpoints, and load their pretrained weights without crashing.

### DIFF
--- a/terratorch/models/backbones/torchgeo_resnet.py
+++ b/terratorch/models/backbones/torchgeo_resnet.py
@@ -74,6 +74,18 @@ look_up_table = {
     "B10": "CIRRUS",
     "B11": "SWIR_1",
     "B12": "SWIR_2",
+
+    "B1": "COASTAL_AEROSOL",
+    "B2": "BLUE",
+    "B3": "GREEN",
+    "B4": "RED",
+    "B5": "RED_EDGE_1",
+    "B6": "RED_EDGE_2",
+    "B7": "RED_EDGE_3",
+    "B8": "NIR_BROAD",
+    "B8a": "NIR_NARROW",
+    "B9": "WATER_VAPOR",
+
     "VV": "VV",
     "VH": "VH",
     "R": "RED",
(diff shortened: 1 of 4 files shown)


romeokienzler:
@clementgilli tnx for ur pr, lvgtm! retriggered the test after rebasing to main (pulling in a fix for one failing test)

clementgilli:
@romeokienzler perfect now ! have a nice day

romeokienzler (approved):

PR #1202, opened 2026-06-30
fix: OlmoEarth backbone forward pass to match olmoearth_pretrain API

whatnick:
## Summary

Fixes the OlmoEarth backbone integration to work correctly with the `olmoearth-pretrain` package API.

## Changes

- **Modality name**: Use `sentinel2_l2a` (matching `MaskedOlmoEarthSample` NamedTuple fields) instead of the invalid `sentinel2`
- **Input format**: Pass raw pixels in `[B, H, W, T, C]` format — the encoder handles patchification internally
- **Mask format**: Use pixel-resolution mask `[B, H, W, T, num_band_sets]` with correct band set count (3 for S2 L2A)
- **Timestamps**: Use `[B, T, 3]` long tensors with `[day, month, year]` (not float zeros)
- **Embed dim**: Read from `encoder.embedding_size` at runtime instead of relying on potentially stale config values
- **Feature extraction**: Use per-modality spatial tokens from `tokens_and_masks` (averaged over T and band_sets), not the global `project_aggregated` vector
- **Config fix**: Corrected Nano embed_dim from 64 → 128

## Testing

All 8 tests in `tests/test_olmoearth.py` pass (~2 min on CPU, no GPU required):
- Registration tests (2)
- Instantiation without weights (1)
- Forward pass shape verification (1)
- Multiple input sizes (1)
- Registry-based creation (1)
- Invalid variant error (1)
- Import guard (1)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - 'feature/**'
     paths-ignore:
       - 'docs/**'
       - '*.md'
@@ -93,6 +94,35 @@ jobs:
       - name: Test with pytest (slow)
         run: |
           pytest -s -v -m "slow" tests
+
+  olmoearth-tests:
+    environment: cicd
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    strategy:
+      matrix:
+        python-version: ["3.13"]
+    env:
+      OMP_NUM_THREADS: "1"
+      MKL_NUM_THREADS: "1"
+      OPENBLAS_NUM_THREADS: "1"
+      NUMEXPR_NUM_THREADS: "1"
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v5
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+      - name: Install dependencies
+        run: |
+          pip install pip==25.1
+          pip --version
+          pip install -e .[test,olmoearth]
+      - name: Test OlmoEarth integration
+        run: |
+          pytest -s -v tests/test_olmoearth.py
 
   rfdetr-tests:
     environment: cicd
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -119,8 +119,8 @@ jobs:
         run: |
           pip install pip==25.1
           pip --version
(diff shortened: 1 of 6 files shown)


romeokienzler:

Thanks @whatnick! Ran on our IBM BlueVela LSF cluster (fresh venv, `pip install -e ".[test]"`, Python 3.12).

**Code assessment.** Fixes the OlmoEarth backbone forward pass to match the `olmoearth_pretrain` API and adds the optional dependency `olmoearth = ["olmoearth-pretrain>=0.1.0"]`. The tests are correctly skip-guarded on that optional dep.

**Unit tests** — `tests/test_olmoearth.py`: **1 passed, 7 skipped** ⚠️

The 7 skips are because `olmoearth-pretrain` is not installed in the base `[test]` extra, so the skip guards fire. That means the substantive forward-pass tests did **not** actually execute in my run — only the registry/import-level test ran. Could you confirm you've run the full suite locally with `olmoearth-pretrain` installed, or point me at the expected results? I'm happy to re-run with the `[olmoearth]` extra installed if that's the intended CI configuration.

**Integration (base set):** not applicable — OlmoEarth is not part of `test_models_fit`.

**Note (unrelated):** 1 pre-existing collection error in `tests/test_geobench_v2_data_module.py` (fixed separately on `main`), not caused by this branch.

romeokienzler:
Hi @whatnick — following up on my review above. 🙏 The PR looks good and is `MERGEABLE`; two small things would let me approve and merge:

1. **The substantive forward-pass tests don't run without the optional dep.** In `tests/test_olmoearth.py` I got **1 passed, 7 skipped** — the 7 skips fire because `olmoearth-pretrain` isn't in the base `[test]` extra, so only the registry/import test actually executed. Could you either:
   - confirm you've run the full suite locally with `olmoearth-pretrain` installed (and share/point me at the results), **or**
   - make CI exercise them (e.g. install the `[olmoearth]` extra in the test job so the forward-pass tests aren't silently skipped)?
   
   I'm happy to re-run on our side with `[olmoearth]` installed if you can confirm that's the intended config.

2. **Just a heads-up, not your bug:** the red `rfdetr-tests (3.13)` check is a known pre-existing/unrelated upstream failure (the rfdetr `lwdetr.py` unpack error), not caused by this branch — no action needed from you there.

Once (1) is sorted I'll approve and merge. Thanks a lot for the fix!

whatnick:
I have applied the proposed extra package install fixes. Please let us know review results.

piperwolters:
Have you considered using the `olmoearth_pretrain_minimal` package instead of `olmoearth_pretrain`? Lighter weight install

PR #1203, opened 2026-07-02
Added temporary fix to support vLLM >0.21.0 and <=0.24.0

christian-pinto:
In vLLM >0.21.0 and <=0.24.0 models that do not register a "language model" fail loading. This pr installs a fix that makes the Terratorch models load and work as expected.

A proper fix is already on the way and will be included in vLLMm > 0.24.0


romeokienzler:
@ahmedemam576 cn u pse have a look why the detr tests r not passing?

romeokienzler (approved):

PR #1204, opened 2026-07-02
Pr 1203

ahmedemam576:
**Problem:**

RF-DETR version got an update.
Some versions expect 2 values from the backbone, newer ones expect 3(Features, Positions, cross-attn-features [new]).
Our code always returned 2, so tests failed with:
expected 3, got 2.

  **Fix:**

I updated terratorch/models/detr/terratorch_detr.py to support both cases[set the cross-attn-features to None and ignore it in the next steps ].
It now retries once and switches to the right format automatically.


christian-pinto (changes requested):
This PR includes code from another PR #1203, please remove that code and push again.

PR #1206, opened 2026-07-10
integrationtests: make /dccstor data paths overridable via TERRATORCH_DATA_ROOT

romeokienzler:
## Summary

The integration tests (`integrationtests/test_base_set.py` + its fit configs) hard-code the shared `/dccstor` storage prefix for input datasets. On clusters where `/dccstor` is not mounted, `test_models_fit` fails immediately with `PermissionError: [Errno 13] Permission denied: '/dccstor'`, which cascades to every dependent predict test.

`TERRATORCH_TMP_ROOT` and `TERRATORCH_TEST_MODELS_ROOT` were already overridable, but the **input-dataset roots** in the fit configs (`/dccstor/geofm-datasets/...`, `/dccstor/geofm-finetuning/...`) were not. This PR closes that gap.

## Changes

- Add `TERRATORCH_DATA_ROOT` (default `/dccstor`). `TEST_MODELS_ROOT` and `TMP_ROOT` defaults now derive from it.
- In `test_models_fit`, after the existing `→ TMP_ROOT` substitution, relocate any remaining `/dccstor` paths onto `DATA_ROOT`, so dataset roots follow the same override.
- Document `TERRATORCH_DATA_ROOT` / `TERRATORCH_TEST_MODELS_ROOT` / `TERRATORCH_TEST_CHECKPOINTS_ROOT` / `TERRATORCH_TMP_ROOT` in the LSF runner README, with a mirror example.

## Backward compatibility

Fully backward-compatible: with the env vars unset, every path resolves to its original `/dccstor` location, so behaviour on the shared cluster is unchanged.

## Testing

Verified the substitution resolves all base-set config dataset/output/model roots onto a local mirror (no bare `/dccstor` left) when `TERRATORCH_DATA_ROOT` points at a mirror that preserves the `/dccstor/...` sub-paths. End-to-end run against a full CCC→BlueVela data mirror in progress.

### DIFF
--- a/integrationtests/test_base_set.py
+++ b/integrationtests/test_base_set.py
@@ -24,11 +24,16 @@ from terratorch.registry import BACKBONE_REGISTRY
 from terratorch.tasks import InferenceTask
 
 # Allow overriding default test asset locations via environment variables.
+# DATA_ROOT is the shared storage prefix that all the default paths below live
+# under. Overriding it (e.g. when the /dccstor mount is unavailable) relocates the
+# input datasets referenced by the fit configs; it defaults to "/dccstor" so the
+# behaviour is unchanged when unset.
+DATA_ROOT = os.getenv("TERRATORCH_DATA_ROOT", "/dccstor")
 TEST_MODELS_ROOT = os.getenv(
     "TERRATORCH_TEST_MODELS_ROOT",
-    "/dccstor/terratorch/shared/integrationtests/testing_models",
+    f"{DATA_ROOT}/terratorch/shared/integrationtests/testing_models",
 )
-TMP_ROOT = os.getenv("TERRATORCH_TMP_ROOT", "/dccstor/terratorch/tmp")
+TMP_ROOT = os.getenv("TERRATORCH_TMP_ROOT", f"{DATA_ROOT}/terratorch/tmp")
 TEST_CHECKPOINTS_ROOT = os.getenv("TERRATORCH_TEST_CHECKPOINTS_ROOT", TEST_MODELS_ROOT)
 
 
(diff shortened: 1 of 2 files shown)


integrationtests/test_base_set.py
```diff
@@ -24,11 +24,16 @@
...
+# input datasets referenced by the fit configs; it defaults to "/dccstor" so the
+# behaviour is unchanged when unset.
+DATA_ROOT = os.getenv("TERRATORCH_DATA_ROOT", "/dccstor")
 TEST_MODELS_ROOT = os.getenv(
     "TERRATORCH_TEST_MODELS_ROOT",
-    "/dccstor/terratorch/shared/integrationtests/testing_models",
+    f"{DATA_ROOT}/terratorch/shared/integrationtests/testing_models",
 )
-TMP_ROOT = os.getenv("TERRATORCH_TMP_ROOT", "/dccstor/terratorch/tmp")
+TMP_ROOT = os.getenv("TERRATORCH_TMP_ROOT", f"{DATA_ROOT}/terratorch/tmp")
```
christian-pinto, line 36: Is {DATA_ROOT} valid here? I can't see where this is replaced with the actual path.

christian-pinto (approved):
Just see my one comment. All good otherwise.

integrationtests/test_base_set.py
romeokienzler: @christian-pinto it's an f-string, it is set above
DATA_ROOT = os.getenv("TERRATORCH_DATA_ROOT", "/dccstor") - so default on CCC, set on bluevela, it is working on bluevela atm

PR #1212, opened 2026-07-27
Security hardening: torch.load(weights_only=True)

adamjstewart:
Follow-up to #391 that patches all remaining locations.


romeokienzler (approved):
Clean, well-scoped security hardening — sets weights_only=True at all remaining torch.load call sites. All changed sites load data consumed as plain state/tensor dicts, so restricting the unpickler is safe. LGTM.

Issue #1199, opened 2026-06-24
Tiled inference API is suboptimal


lnicola:
**Is your feature request related to a problem? Please describe.**
The `tiled_inference` API requires the input and output tensors to be in memory. This makes it very hard to use on really large inputs.

**Describe the solution you'd like**
A more flexible API that lets the user plug in the read part (source) and the output processing (sink). This makes it possible to read the input on demand and write the output right away instead of storing everything in memory.

**Describe alternatives you've considered (optional)**
As a workaround, the user can slice the input and call `tiled_inference` multiple times, taking care to overlap and crop the "super-tiles". I implemented this, but it wasn't easy.

Note that it's useful for the "super-tiles" to be larger than the model patch size, since they seem to affect the inference performance, possibly due to batching or some other Torch particularities.

romeokienzler:
Hi @lnicola thanks for the issue, you think there is a chance you submit a PR fixing this?

lnicola:
Sorry, not feeling confident enough to try my hand at it (especially since I didn't manage to make inference deterministic), but I can at least review a PR that implements it.

akshanthsaik:
@romeokienzler , we just had a brief chat on Slack, and I’d like to take this up.I’d like to hear your and @lnicola 's expectations and how you’d like me to approach it.

Should we keep the existing tiled_inference function and add a new one, or refactor the existing one?
Are there existing source/sink patterns in the codebase I should follow?

Thanks!

guillemc23:
@lnicola we are using [SAHI](https://obss.github.io/sahi/) to perform tiled inference, the implementation with a Terratorch model is not hard, you might want to check it out 🤓

bhansaliraunak:
@lnicola To answer your question based on here https://github.com/torchgeo/terratorch/pull/1213#issuecomment-5150303963

**Does prepare_tiled_inference_input still require an input_batch tensor?**
Currently it does. Looking at `prepare_tiled_inference_input()` and `get_input_chips()`:

- It expects input_batch to be either a torch.Tensor or dict
- It then reads input_batch.shape to determine input_batch_size, h_img, w_img, and device.

In get_input_chips, it slices patches directly from input_batch and attaches that patch as InferenceInput.input_data

So even if a custom `TileSource` is provided, `prepare_tiled_inference_input` still requires an `input_batch` tensor to derive the dimensions and batch size.

bhansaliraunak:
[...]

PR #1213, opened 2026-07-29
feat: add pluggable source sink api for out of core inference

bhansaliraunak:
Introduces two optional Protocol-based plug-in points to tiled_inference()
that decouple chip reading and prediction writing from the in-memory tensor
path, enabling inference on images too large to fit in RAM.

- TileSource: callable protocol; return a chip tensor on demand (e.g. from
  a rasterio window) instead of reading from a pre-loaded tensor
- TileSink: write() + finalize() protocol; stream predictions directly to
  disk instead of accumulating them in a large in-memory buffer
- InMemoryTileSource / InMemoryTileSink: default implementations that
  reproduce the original behaviour exactly so no breaking changes
- input_batch is now optional (None allowed when a custom source is used)
- tiled_inference() return type broadened to Tensor | None to support sinks
  that write to external stores and return None from finalise()
  
  #1199

### DIFF
--- a/terratorch/tasks/tiled_inference.py
+++ b/terratorch/tasks/tiled_inference.py
@@ -3,12 +3,35 @@ This does some additional things over the obvious fold -> predict -> unfold logi
 e.g. cropping out areas around model prediction to reduce artifacts
 
 It additionally rebatches after the fold operation to gain speed up.
+
+Source / Sink API
+-----------------
+The :func:`tiled_inference` function accepts optional *source* and *sink* objects that
+decouple **reading** and **writing** from the in-memory tensor path.
+
+* A **source** is any callable that matches :class:`TileSource`.  Given an
+  :class:`InferenceInput` descriptor it should return the chip tensor
+  ``(C, H, W)`` – or ``(C, T, H, W)`` for multi-temporal inputs – on the CPU.
+  This lets callers read from disk (e.g. a rasterio window) instead of holding
+  the whole image in memory.
+
+* A **sink** is an object that matches :class:`TileSink`.  Its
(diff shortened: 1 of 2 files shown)


lnicola:
I'm having some trouble following the code, but does `prepare_tiled_inference_input` still require an `input_batch` tensor?

Also, having the sink do the blending of patches makes implementing a sink hard. I would have hoped to let the library do the hard work of overlapping and cropping and blending.

romeokienzler:

Thanks @bhansaliraunak! I ran this branch on our IBM BlueVela LSF cluster (fresh venv, `pip install -e ".[test]"`, Python 3.12).

**Code assessment.** The pluggable Source/Sink Protocol API for out-of-core tiled inference is a clean, well-scoped addition — the abstraction is self-contained and the new code path is exercised by a dedicated test module.

**Unit tests** — `tests/test_tiled_inference.py`: **17 passed** ✅

**Integration (base set).** Not applicable here — `integrationtests/test_base_set.py::test_models_fit` does not exercise the tiled-inference path, so I did not run it against this PR. The new functionality is covered by the unit tests above.

**Note (unrelated to this PR):** the full-tree `pytest --collect-only` shows 1 collection error in `tests/test_geobench_v2_data_module.py`. This is **pre-existing on `main`** (an optional-dependency import guard, being fixed separately) and is not caused by this branch.

LGTM from the test side. 👍

romeokienzler (approved):
Approving after maintainer review + full unit suite (and integration base set where applicable) on the IBM BlueVela cluster — all green. The failing CI check on this PR is a stale/unrelated pre-existing failure (geobench_v2 optional-dep collection error / upstream rfdetr incompat), not caused by this branch.

PR #1218, opened 2026-08-14
Add GFM Swin Encoder

nilsleh:
Adds the GFM Swin-B encoder, an RGB only model,  to the backbone registry, usable from EncoderDecoderFactory 

- Paper:  [Towards Geospatial Foundation Models via Continual Pretraining](https://arxiv.org/abs/2302.04476) (arXiv:2302.04476)  
- Repo: https://github.com/mmendiet/GFM 
- Weights: [torchgeo/gfm](https://huggingface.co/torchgeo/gfm) (Apache-2.0)

- Registers gfm_swin_base, wrapping MMSegSwinTransformer to return a four-level BCHW pyramid — channels [128, 256, 512, 1024] at strides [4, 8, 16, 32].
- No PermuteDims neck required, unlike satlas_swin_* and prithvi_swin_* which emit NHWC — verified with UperNetDecoder and UNetDecoder.
- checkpoint rehosted at https://huggingface.co/torchgeo/gfm
- The 8 missing per-stage LayerNorms are harmless. mmseg's Swin has a norm per stage that the GFM checkpoint does not carry, so they stay at PyTorch's default init — weight=1, bias=0,

### DIFF
--- a/docs/package/backbones.md
+++ b/docs/package/backbones.md
@@ -30,6 +30,27 @@
     options:
         toc_label: "Clay v1"
 
+:::terratorch.models.backbones.gfm_swin.gfm_swin_base
+    options:
+        toc_label: "GFM Swin"
+
+!!! tip
+    `gfm_swin_base` returns NCHW features, so unlike the other Swin backbones it needs no
+    `PermuteDims` neck to feed decoders such as `UperNetDecoder`:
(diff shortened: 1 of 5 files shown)


nilsleh:
waiting for #1221

PR #1219, opened 2026-08-17
Fix DDP deadlock with CombinedLoss (#982)

### LINKED ISSUE #982: combined losses for segmentation task break distributed training
**Describe the issue**
I installed the most recent terratorch version from the main branch (commit `7cbf469f1c179860b81117512f873c6ff2b1ff5a`), to use the latest features (specifically, I was interested in combining multiple losses for semantic segmentation, like "ce"+"dice"). I used:
pip install git+https://github.com/IBM/terratorch.git

However, with this version installed, trainer.fit() hangs forever after the first train+val epoch. This only happens when **distributing training** over more than 1 GPUs and when using a **CombinedLoss**. It works when using a single loss, like "dice", instead.

**To Reproduce**
I attach a simple script for semantic segmentation trained with random data, to reproduce the issue:
```
import lightning as L
from terratorch.tasks import SemanticSegmentationTask
from torch.utils.data import Dataset, DataLoader
import torch

LOSS = "dice"  # works
# LOSS = {"ce": 1.0, "dice": 5.0}  # does not work


class DummyDataModule(L.LightningDataModule):
    def setup(self, stage) -> None:
        self.train_dataset = RandomDataset()
        self.val_dataset = RandomDataset()

    def train_dataloader(self) -> DataLoader:
        return DataLoader(
            dataset=self.train_dataset,
            batch_size=1,
            shuffle=True,
            num_workers=2,
            pin_memory=True,
            drop_last=True,
        )

    def val_dataloader(self) -> DataLoader:
        return DataLoader(
            dataset=self.val_dataset,
            batch_size=1,
            shuffle=False,
            num_workers=2,
            pin_memory=False,
            drop_last=False,
        )


class RandomDataset(Dataset):
    def __init__(self):
        self.imgs = [torch.rand((4, 4, 64, 64)) for _ in range(100)]
        self.masks = [torch.ones((64, 64)).long() for _ in range(100)]
        super().__init__()

    def __getitem__(self, idx):

        return {
            "image": self.imgs[idx],
            "mask": self.masks[idx],
        }

    def __len__(self):
        return len(self.imgs)


module = SemanticSegmentationTask(
    model_factory="EncoderDecoderFactory",
    model_args={
        "backbone": "prithvi_eo_v2_300",
        "backbone_pretrained": True,
        "backbone_img_size": 64,
        "backbone_num_frames": 4,
        "backbone_bands": ["BLUE", "GREEN", "RED", "NIR_NARROW"],
        "backbone_coords_encoding": [],
        "necks": [
            {"name": "SelectIndices", "indices": [5, 11, 17, 23]},
            {"name": "ReshapeTokensToImage", "effective_time_dim": 4},
            {"name": "LearnedInterpolateToPyramidal"},
        ],
        "decoder": "UNetDecoder",
        "decoder_channels": [512, 256, 128, 64],
        "head_dropout": 0.1,
        "num_classes": 2,
    },
    loss=LOSS,
    lr=1e-4,
    optimizer="AdamW",
    ignore_index=-1,
    freeze_backbone=False,
    freeze_decoder=False,
    plot_on_val=False,
)
datamodule = DummyDataModule()
trainer = L.Trainer(
    max_epochs=5,
    default_root_dir="./logs",
    num_sanity_val_steps=2,
    precision="bf16-mixed",
    log_every_n_steps=10,
    limit_val_batches=5,
    limit_train_batches=5,
)
trainer.fit(module, datamodule=datamodule)
```

**Screenshots or log output**
Lightning fit progress bar simply gets stuck at the end of the first epoch:
Image

**Expected behavior**
Training should work with CombinedLoss, too. 

**Deployment information**
Describe what you've deployed and how:
 - TerraTorch version: latest version from main branch (commit 7cbf469f1c179860b81117512f873c6ff2b1ff5a)
 - Installation source: pip
 - torch version: 2.9.0 (cu12)
 - devices: 2x NVIDIA TITAN RTX GPUs

I add @blumenstiel, since I noticed that the CombinedLoss feature has been recently added by a PR from you. Sorry if you are not the right person to ask.

romeokienzler:
## Summary

Fixes #982 — `trainer.fit()` hangs at the end of the first epoch when training with a `CombinedLoss` (e.g. `{"ce": 1.0, "dice": 5.0}`) across **more than one GPU**. A single loss works fine.

## Root cause

In `LossHandler.log_loss`, the **total** loss is logged with Lightning's default `on_step`/`on_epoch` behaviour (context-dependent), but each **individual component** loss was logged with `on_step=True, on_epoch=True` forced, together with `sync_dist=True`:

```python
log_function(f"{self.loss_prefix}loss", full_loss, sync_dist=True, batch_size=batch_size)  # defaults
for loss_name, loss_value in all_losses.items():
    log_function(..., on_epoch=True, on_step=True, sync_dist=True, ...)  # forced both
```

The forced epoch-level **synchronized** reduction fired only for the component losses and not for the total loss. That asymmetry desynchronised the DDP collective calls across ranks, deadlocking at the first epoch boundary. Single losses never enter the component loop, which is exactly why only `CombinedLoss` was affected.

This matches the debugging by @matteot11 and @bhansaliraunak in the issue thread.

## Fix

Log the component losses with the **same** settings as the total loss — letting Lightning pick `on_step`/`on_epoch` from the hook context (`on_step` during training, `on_epoch` during validation/test), keeping `sync_dist=True`. This is the recommended Lightning behaviour and keeps the collectives symmetric across ranks.

Note on @blumenstiel's point about keeping both step- and epoch-level monitoring: the default per-hook behaviour already gives per-step curves during training and epoch-averaged values during validation, which is the usual monitoring setup. Forcing both on the components is precisely what deadlocks, so it can't be retained as-is.

## Tests

Adds `tests/test_loss_handler.py` — a regression guard asserting that neither the total nor the component losses force `on_step`/`on_epoch`, that logging doesn't mutate the caller's dict, and that `CombinedLoss` component keys line up with what gets logged.

## Validation

Ran on the IBM BlueVela LSF cluster (Python 3.12): **`tests/test_loss_handler.py` — 4 passed**.

_Note: a true 2-GPU hang/no-hang reproduction of the original script would be the ideal end-to-end check; I validated the logging contract that caused the desync. Reviewers with a multi-GPU box can confirm the original repro from the issue now completes._

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/terratorch/tasks/loss_handler.py
+++ b/terratorch/tasks/loss_handler.py
@@ -93,12 +93,17 @@ class LossHandler:
         full_loss = all_losses.pop("loss")
         log_function(f"{self.loss_prefix}loss", full_loss, sync_dist=True, batch_size=batch_size)
 
+        # Log the individual component losses (present when using a CombinedLoss) with the same
+        # settings as the total loss above, i.e. letting Lightning pick on_step/on_epoch from the
+        # hook context (on_step during training, on_epoch during validation/test). Forcing both
+        # on_step=True and on_epoch=True here previously caused DDP to deadlock at the end of the
+        # first epoch, because the extra epoch-level synced reduction was only triggered for the
+        # component losses and not for the total loss, desynchronising the collective calls across
+        # ranks. See https://github.com/IBM/terratorch/issues/982.
         for loss_name, loss_value in all_losses.items():
             log_function(
                 f"{self.loss_prefix}{loss_name}",
                 loss_value,
-                on_epoch=True,
-                on_step=True,
                 sync_dist=True,
                 batch_size=batch_size,
             )
--- a/tests/test_loss_handler.py
+++ b/tests/test_loss_handler.py
@@ -1,0 +1,75 @@
+# Copyright contributors to the Terratorch project
+"""Tests for terratorch.tasks.loss_handler.
+
+Regression coverage for https://github.com/IBM/terratorch/issues/982: logging the individual
+component losses of a CombinedLoss must not force ``on_step``/``on_epoch``. Doing so made DDP
+deadlock at the end of the first epoch, because the extra epoch-level synced reduction fired only
+for the component losses (and not for the total loss), desynchronising the collective calls across
+ranks.
+"""
+import torch
+
+from terratorch.tasks.loss_handler import CombinedLoss, LossHandler
+
+
+def _capture_logs(loss_dict, prefix="train/"):
+    """Run log_loss with a fake log_function and return the recorded calls."""
+    calls = []
+
+    def fake_log(name, value, **kwargs):
+        calls.append({"name": name, "value": value, "kwargs": kwargs})
+
+    handler = LossHandler(prefix)
+    handler.log_loss(fake_log, loss_dict=loss_dict, batch_size=4)
+    return calls
+
+
+def test_single_loss_logs_only_total():
+    calls = _capture_logs({"loss": torch.tensor(1.0)})
+    assert [c["name"] for c in calls] == ["train/loss"]
+    # No on_step/on_epoch forced; Lightning defaults are used.
+    assert "on_step" not in calls[0]["kwargs"]
+    assert "on_epoch" not in calls[0]["kwargs"]
+    assert calls[0]["kwargs"]["sync_dist"] is True
+
+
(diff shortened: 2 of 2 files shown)


rosielickorish (approved):
LGTM

PR #1221, opened 2026-08-18
Update terratorch to torchgeo 0.10

nilsleh:
The latest release of torchgeo 0.10 causes some deprecation issues. These are addressed in this PR such that the CLI tests can pas again.

- torchgeo.trainers changed to torchgeo.task
- quantile_normalization became percentile_normalization

Edit: PR also tries to fix a bunch of other failing CI tests:

- Added our own unbind_samples in terratorch/datasets/utils.py that leaves non-tensors alone, like torchgeo ≤0.9 did.
- Pointed the seven task modules at it instead of torchgeo's.
- Widened the plotting guards to catch TypeError too, because a failed debug plot should not kill the entire run
- tacoreader was missing from test environment and causing failing tests


terratorch/datamodules/geobench_v2_data_module.py
```diff
@@ -37,7 +37,7 @@
 from matplotlib.figure import Figure
 from torch import Tensor, nn
 from torch.utils.data import default_collate
-from torchgeo.datasets.utils import lazy_import, percentile_normalization
+from torchgeo.datasets.utils import lazy_import, quantile_normalization
```
adamjstewart, line 40: Neither of these are public. percentile_normalization will be removed in 1.0, and quantile_normalization will be moved to a classmethod in 0.11. Only publicly documented functions from TorchGeo should be used.

terratorch/tasks/base_task.py
```diff
@@ -8,9 +8,9 @@
 import pandas as pd
 import torch
 from lightning.pytorch.callbacks import Callback
-from torchgeo.datasets.utils import unbind_samples
-from torchgeo.trainers import BaseTask
+from torchgeo.tasks import BaseTask
 
+from terratorch.datasets.utils import unbind_samples
```
adamjstewart:
```suggestion
from terratorch.datasets import unbind_samples
```
This one is public if you import it correctly

nilsleh:
image

calebrob6:
@romeokienzler can you approve this? We need to sort out the compatibility with the latest torchgeo release as currently we can't use torchgeo 0.10 and terratorch together.

Issue #912, opened 2025-10-01
Testing environment is not replicable outside of CI


christian-pinto:
There is no way for the exact testing environment used in CI to be replicated locally, if not by copy-pasting commands from the github workflows configuration files. This is especially true for the integration tests where it is not clear which environment the tests are running in.

For replicability purposes, tests should be executed in a sand-boxed python environment programmatically created for the tests and then deleted. This would make the tests completely oblivious to the local environment and any of its potential quirks. e.g., someone fiddling with `PYTHONPATH`, forgetting about it and then debugging test failures that might not necessarily depend on their code but rather on the state of their python env.

Joao-L-S-Almeida:
Hi, @christian-pinto . We are starting to deal with it in the experimental branch `containerize_tests`.

romeokienzler:
@christian-pinto 

in the project's root run:
pip install . .[test]
pytest tests
pytests integrationtests/test_baseset.py

tests are running in gh wf, integrationtests on ccc atm using a custom cicd script but we r transitioning to a self hosted gh runner on ccc or zuvela atm....

romeokienzler:
@romeokienzler close one done
@christian-pinto cn u pse confirm that it is working and u r unblocked?

christian-pinto:
@romeokienzler i am not blocked because I run the vLLM integration tests manually before pushing a PR. Looking forward to having all tests executed with the gpu powered runner.

Joao-L-S-Almeida:
@romeokienzler @christian-pinto Consider #994 as an alternative to perform it.

romeokienzler:
A concrete instance of this non-reproducibility just bit CI: `torchgeo` was declared **unbounded** in `pyproject.toml` (`"torchgeo"`), so a from-scratch `pip install -e ".[test]"` resolves the latest release. `torchgeo` 0.10.0 removed `torchgeo.trainers.utils` (imported in `terratorch/models/timm_model_factory.py`), so the install dies at import before any test runs:

```
ImportError: cannot import name 'utils' from 'torchgeo.trainers'
```

**Repro:**
```bash
python3.12 -m venv /tmp/tt && source /tmp/tt/bin/activate
pip install -e ".[test]"        # resolves torchgeo 0.10.0
python -c "import terratorch"    # ImportError as above
```

[...]

PR #1222, opened 2026-08-18
Pin torchgeo to >=0.9.0,<0.10 to fix fresh-install import break (#912)

romeokienzler:
## What

Pins `torchgeo` to `>=0.9.0,<0.10` in `pyproject.toml` (it was previously declared unbounded as `"torchgeo"`).

## Why

`torchgeo` **0.10.0** removed `torchgeo.trainers.utils`, which terratorch imports in [`terratorch/models/timm_model_factory.py`](https://github.com/torchgeo/terratorch/blob/main/terratorch/models/timm_model_factory.py) (`from torchgeo.trainers import utils`), along with other breaking API changes. Because the dependency had no upper bound, a from-scratch install resolves the latest (0.10.0) and dies at import before any test runs:

```
ImportError: cannot import name 'utils' from 'torchgeo.trainers'
```

This is what makes the `build (3.13)` and `rfdetr-tests (3.13)` CI jobs red on recent PRs (27 collection errors on the former, 1 on the latter), and it prevents anyone from reproducing the test environment locally per #912.

## Verification

- `torchgeo 0.9.0` satisfies the new spec and `from torchgeo.trainers import utils` imports cleanly.
- The `<0.10` bound excludes the broken 0.10.0.
- The full unit suite (**1603 passed**) and the integration base set (**16 passed**) were run against 0.9.0 on the BlueVela cluster while validating #982/#1219.

Refs #912.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,10 @@ dependencies = [
   "lightning>=2.6.0",
   "segmentation-models-pytorch>=0.5.0",
   "jsonargparse>=4.40.0",
-  "torchgeo", #>=0.7.0,<0.7.2",
+  # torchgeo 0.10 removed torchgeo.trainers.utils (imported in timm_model_factory.py)
+  # and made other breaking API changes; 0.9.x is the verified-compatible line. Without
+  # an upper bound a fresh install resolves 0.10 and fails at import. See issue #912.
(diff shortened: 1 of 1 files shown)


rosielickorish (approved):
LGTM

PR #1239, opened 2026-09-11
Security hardening: weights_only=True

adamjstewart:
Follow-up to #391 and #1212. Let me know if you want to update https://github.com/torchgeo/terratorch/security/advisories/GHSA-6qgg-frm7-fc93 to include 1.2.13 or if I should open a new report. We need to stop this kind of thing from entering the code base.


calebrob6:
@adamjstewart would we be able to rule out these sort of security issues by policy? If I'm understanding correctly, to trigger a potential remote code execution, a user needs to call `load_dofa_weights(..., weights=some_malicious_file_they_downloaded)`

adamjstewart:
Correct, a security policy that says "TerraTorch is not secure by design" would certainly close most security issues. However, TerraTorch is already widely used inside NASA and I would love to bring its FMs to ESA servers as well. Since it's easy to close security holes, I don't see a good reason not to.

PR #1240, opened 2026-09-14
Enable support for Terratorch models in v2 engine

christian-pinto:
#### New: `terratorch/vllm/model_state.py` *(new file, +93 lines)*
Introduces `TerratorchModelState`, a custom subclass of vLLM's `DefaultModelState` for the V2 engine runner. Terratorch models pass raw multimodal tensors directly to `forward()` rather than through the standard encoder-embedding path. This class handles:
- **`prepare_inputs()`** — extracts real multimodal tensors from the encoder cache for live requests; falls back gracefully to dummy tensors during warmup/profile passes when no features are present.
- **`prepare_dummy_inputs()`** — synthesises correctly-shaped `(1, C, T, H, W)` tensors (matching what `group_and_batch_mm_kwargs` produces) so warmup passes receive the expected input shape.
- **`prepare_inputs_embeds()` / `dummy_inputs_embeds()`** — both return `None`, since Terratorch bypasses the `inputs_embeds` path entirely.

#### Modified: `terratorch/vllm/__init__.py`
Exports `TerratorchModelState` from the `terratorch.vllm` package so downstream code can import it directly.

#### Modified: `segmentation_io_processor.py`
- Removed the runtime `check_vllm_version` guard; multimodal data is now always wrapped as `{"image": multi_modal_data}` (correct for all supported vLLM versions ≥ 0.14.0).
- Added a `None` guard for `location_coords` — only included in the payload if the config requests it **and** the value is not `None`.

#### Modified: `terramind_segmentation_io_processor.py`
- Same version-check removal as above — always wraps multimodal data in `{"image": ...}`.
- Updated comment to explain why the version check was removed (unreliable for dev installs).

---

### Summary
The primary addition is a custom `ModelState` that lets Terratorch models work correctly with the vLLM V2 worker. The supporting changes clean up fragile version-check logic in both IO processors, replacing conditional wrapping with the single always-correct code path.


romeokienzler (approved):
Code review: The changes are clean and well-scoped to the vllm module. The new `TerratorchModelState` correctly handles warmup/profile passes via dummy tensor synthesis. The version-check removal is justified — the removed conditional was unreliable for dev installs and the always-correct path is now used. The `location_coords` None guard in `segmentation_io_processor.py` is a good defensive fix. The only failing CI check (`build (3.13)`) reproduces identically on the current `main` branch (same 2 `test_clay_v15_backbone.py` precision failures) and is unrelated to this PR. Approving.

